PR #8, opened 2025-09-04
[Feature] support Ascend 310P

richard4fan:
add initial support for Ascend 310P NPU

### DIFF
--- a/lmcache_ascend/v1/memory_management.py
+++ b/lmcache_ascend/v1/memory_management.py
@@ -25,7 +25,7 @@ def is_310p():
     global _IS_310P
     if _IS_310P is None:
         from lmcache_ascend import _build_info
-        _IS_310P = _build_info.__soc_version__.lower().startwith("ascend310p")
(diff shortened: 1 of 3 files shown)


lmcache_ascend/v1/memory_management.py
```diff
@@ -19,6 +19,15 @@
 
 logger = init_logger(__name__)
 
+_IS_310P = None
+
+def is_310p():
+    global _IS_310P
+    if _IS_310P is None:
+        from lmcache_ascend import _build_info
+        _IS_310P = _build_info.__soc_version__.lower().startwith("ascend310p")
```
matthewygf, line 28: Should this be .startswith ??

matthewygf (commented):
In general, looks good ! Just one comment on the startwith

lmcache_ascend/v1/memory_management.py
richard4fan: yes , my fault, i will fix it today and do a test again.

PR #14, opened 2025-09-12
Update causal in flash_attn_varlen_func

zahrayousefijamarani:
causal param should be True to generate correct answers during inference.


matthewygf:
@zahrayousefijamarani Thanks for the contribution.

While you are correct that, the causal param should be true. The related [issue](https://github.com/LMCache/LMCache/issues/1152) seems to indicate that without the sparse attention impl, causal = true will lead the perf degradation. I am a bit hesitate that we turn this on without the sparse attn impl. WDYT ?

cc @egretwAlker @YaoJiayi

PR #65, opened 2025-11-20
Fix multi kernel v2 to tile by tokens, avoid OOM

matthewygf:
## Background
The existing v1 kernel uses all available AIVs which makes overlap computation and communication sometimes unavailable.
The proposed previous v2 kernel version can reduce the AIVs by 10x while achieving similar performance (~10% degradation, in the grand scheme of us, that is fine)

However, the previous v2 version did not tile the tokens properly and therefore can lead to OOM over the UB.

## Fix
This current version fixes the OOM by tile the tokens within the UB size.


matthewygf:
image
Multi layer kernels specific unit tests passes

PR #108, opened 2025-12-25
add MooncakeStoreConnector support for mindspore

liwei-rk:
make numpy.ndarray compatible for pytorch tensor in mooncake store.


lmcache_ascend/mindspore/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -0,0 +1,182 @@
...
+            return results
+
+        except Exception as exc:
+            logger.error(f"batch_get_into threw exception: {str(exc)}")
+            # Release any buffers we successfully allocated
+            for i in valid_idx:
+                memory_objs[i].ref_count_down()  # type: ignore
+            return [None] * len(keys)
+
+async def MooncakeStoreConnector__batched_put_zero_copy(
```
chloroethylene: no this function in v0.3.7

chloroethylene:
it would be better to construct new object and rewrite interface like buffer.data_ptr(). do it later.

Issue #60, opened 2025-11-18
Staging Copy C++ Operator instead of ZeroCopy / NativeTorch copy + kernel transfer


matthewygf:
## Background

Currently LMCache provides two way of copying kvcaches, zero-copy with disabling use_gpu_buffer, or via a staging buffer using tensor.copy_.

On A3 machines with Aarch64, zero-copy can achieves high bandwidths from ~30GB/s to 50GB/s depends on number of AIV. On A2 machines with Aarch64, zero-copy can achieves bandwidths ranging from 7GB/s to 12GB/s.

However, on x86 architecture, A2 machine bandwidths drop to ~5GB/s.

## Potential Improvements

Using memcpy + kernel can sometimes achieve higher bandwidth the zero-copy on x86 architecture.

However, using tensor.copy_ + lmc_ops.xxxtransfer can introduce host overhead.

Hence, it can be beneficial to introduce a new operator that fuses the memcpy + kernel transfer into a pybind11 call.

matthewygf:
done via #110

PR #110, opened 2025-12-26
Implementing Staging Copy C++ and fix KVCache format detection bug when using MLA

### LINKED ISSUE #111: Fix KVCache format detection when using MLA
## Bug Description
The current `KVCacheFormat.detect` method fails to correctly determine the KVCache format when using MLA on Ascend NPU. This causes issues in logic that relies on format detection in classes such as `VLLMPagedMemNPUConnectorV2`.

### Example (from current implementation):
```python
class KVCacheFormat(Enum):
    ...
    @staticmethod
    def detect(
        kvcaches: List[Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]]
    ) -> 'KVCacheFormat':
        if not kvcaches:
            return KVCacheFormat.UNDEFINED
        
        first_cache = kvcaches[0]
        
        if isinstance(first_cache, tuple):
            return KVCacheFormat.SEPARATE_KV
        elif isinstance(first_cache, torch.Tensor):
            if first_cache.shape[0] == 2:
                return KVCacheFormat.MERGED_KV
        
        return KVCacheFormat.UNDEFINED
```

This detection logic does not work correctly when KV caches are constructed for MLA, leading to `KVCacheFormat.UNDEFINED` and exceptions in `_initialize_pointers`.

## Expected Behavior
KVCache format should be detected correctly regardless of if MLA is enabled or not, and the code should not enter the undefined state for valid MLA cache formats.

## Steps to Reproduce
1. Use MLA-enabled configuration on Ascend NPU.
2. Call logic that triggers `_initialize_pointers` or `KVCacheFormat.detect` with MLA-style kvcaches.
3. Observe ValueError due to undetected KV cache format.

## Environment
- LMCache-Ascend
- Ascend NPU with MLA

## Possible Solution
- Update `KVCacheFormat.detect` to properly check for MLA-style shapes or add additional detection for NPU/MLA cache structures (e.g., `[num_pages, page_size, hidden_dims]`).
- Update `_initialize_pointers` to properly obtain mla layers pointers.

---

shengyan-chen:
## Changes
- Performance Optimization: Replaced NativeTorch copy + kernel transfer with a Staging Copy (C++ host-side memcpy) strategy. 
- Bug Fix: Fixed the KV Cache format detection logic specifically when using MLA

## Fix and Related issue
- Related to #60 
- Fixes #111 

## Testing
Verified functionality and performance on:
- vLLM 0.10.2: Tested with both multi_layer and layerwise modes.
- vLLM 0.11.0rc1: Tested with multi_layer mode. (not support layerwise now)

## ToDo
- [ ] Add support for Ascend 310P.
- [ ] Adapt layerwise mode for vLLM 0.11.x.

### DIFF
--- a/csrc/mem_kernels.cpp
+++ b/csrc/mem_kernels.cpp
@@ -38,94 +38,105 @@ void multi_layer_kv_transfer(
     const torch::Device &paged_memory_device, const int page_buffer_size,
     const bool direction, const bool use_mla, const int kvcache_format_raw) {
   uint8_t *key_value_ptr = get_kernel_ptr<uint8_t, torch::Tensor>(key_value);
-  // it is actually a uint8_t**. we will reinterpret it inside the kernel
-  uint8_t *page_buffer_ptrs =
-      get_kernel_ptr<uint8_t, const torch::Tensor>(key_value_ptrs);
-  uint8_t *slot_mapping_ptr =
-      get_kernel_ptr<uint8_t, const torch::Tensor>(slot_mapping);
 
-  int num_layers = key_value.size(1);
-  int num_tokens = slot_mapping.size(0);
-  int hidden_dims = key_value.size(-1);
-  int kv_size = 2;
-  if (use_mla) {
-    kv_size = 1;
-  }
+  MultiLayerKVConfig config = prepare_multi_layer_kv_config(
+      key_value, key_value_ptrs, slot_mapping, paged_memory_device,
+      page_buffer_size, direction, use_mla, kvcache_format_raw);
 
-  kvcache_ops::KVCacheFormat kvcache_format =
-      static_cast<kvcache_ops::KVCacheFormat>(kvcache_format_raw);
-
-  const c10::OptionalDeviceGuard device_guard(paged_memory_device);
-  // we require the kv ptr list to be on the device too
-  const c10::OptionalDeviceGuard kv_device_guard(device_of(key_value_ptrs));
-
-  const aclrtStream stream = c10_npu::getCurrentNPUStream().stream();
-  at::ScalarType scalar_type = key_value.scalar_type();
-  at::ScalarType slot_type = slot_mapping.scalar_type();
-  const char *socName = aclrtGetSocName();
-  auto ascendcPlatform =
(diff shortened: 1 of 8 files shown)


csrc/mem_kernels.cpp
```diff
@@ -272,6 +411,243 @@ void single_layer_kv_transfer(torch::Tensor& lmc_key_value_cache,  // [num_token
...
+) {
+
+    size_t num_chunks = lmc_tensors.size();
+    if (chunk_offsets.size() != num_chunks || chunk_sizes.size() != num_chunks) {
+        PyErr_SetString(PyExc_RuntimeError, "chunk_offsets and chunk_sizes must have the same size as lmc_tensors");
+        throw py::error_already_set();
+    }
+
+    if (num_chunks == 0) {
+        return;
```
matthewygf: Do you think we should return a valueError here where num_chunks shouldn't be 0 ?

shengyan-chen: num_chunks == 0 shouldn't be an error, just return directly without proceeding with further operations, What do you think?

csrc/mem_kernels.cpp
```diff
@@ -272,6 +411,243 @@ void single_layer_kv_transfer(torch::Tensor& lmc_key_value_cache,  // [num_token
...
+    int64_t vllm_buffer_size = static_cast<int64_t>(vllm_key_value_cache.nbytes());
+    int64_t staging_buffer_size = static_cast<int64_t>(staging_cache.nbytes());
+
+    int64_t element_size = staging_cache.element_size();
+    int64_t hidden_dim = num_heads * head_dims;
+
+    int64_t bytes_per_token = staging_token_stride * element_size;
+    
+    // use for token_major = false
+    int64_t staging_v_plane_offset = staging_value_offset * element_size;
```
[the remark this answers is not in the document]
shengyan-chen: Alright, I have extracted some of the code into functions.

csrc/mem_kernels.cpp
shengyan-chen: I think you are right. I have modified the code to throw an error when num_chunks == 0.

PR #157, opened 2026-01-27
[feature] LMCache-Ascend 0.3.12 support mindspore

chloroethylene:
LMCache-Ascend 0.3.12 support mindspore


matthewygf (approved):
Minor nit: if you can attach a small notes on why patching storagemanager that will be great! 

lgtm !

PR #172, opened 2026-02-12
Move the log info to right code block

xleoken:
Move the log info to right code block, local variable 'layer_id' can't be referenced before assignment.


matthewygf:
> * Commits must have verified signatures.

I think it is still unverified. you might need to check ```gpgsign``` @xleoken

xleoken:
hi @matthewygf, the commit had been signed.
image

chloroethylene (approved):
lgtm

PR #179, opened 2026-02-25
Add .idea to gitignore

xleoken:
Add `.idea` to .gitignore file, because some developers may use pycharm ide.


xleoken:
thanks @matthewygf

PR #183, opened 2026-02-27
Copy disagg_proxy_server.py to lmcache-ascend project

xleoken:
We don't need to clone lmcache project when the lmcache was installed by using cmd `NO_CUDA_EXT=1 pip install lmcache==0.3.12`. Copy the disagg_proxy_server.py file to lmcache-ascend project will be friendly for users, and it also facilitates debugging for developers.


xleoken:
cc @matthewygf take a review when free, thanks

PR #192, opened 2026-03-17
Migrate HCCL agent to CANN 8.5

matthewygf:
This pull request port HCCL agent to CANN8.5. The HCCL Agent is much lighter weight than both HCOMM and HIXL. Therefore the connection time is much faster.

**Build System and Dependency Management:**

* The CMake and `setup.py` build logic now conditionally includes the `hccl` subdirectory and related dependencies only if `HCOMM_SRC_PATH` is defined, enabling builds with or without internal HCOMM headers. This avoids unnecessary dependencies and makes the build process more flexible. 
  * **However this is a breaking change to older version of lmcache-ascend**

**Test Infrastructure and Usability:**

* Added unittests for HCCL Channel

**Code Cleanup and Refactoring:**

* The initialization logic in `HcclAgent` is updated to correctly pass logical and physical device IDs, and to provide better error handling and debugging output.


matthewygf:
image

matthewygf:
image

PR #231, opened 2026-05-08
fix ci certificate verification failed

jianghv:
When using the standard image, certificate authentication fails during git fetch. It is necessary to use the OpenEuler version certificate from the host machine.


jianghv:
image
Fetch works normally with the host certificate.

chloroethylene (approved):

PR #237, opened 2026-05-15
Adapter Agentos:Add chunk_hash return and expose evict/prefetch APIs

jianghv:
To interface with the upper-layer AgentOS, each KV Cache needs a unique identifier. Therefore, a chunk_hash field is added to the response returned by the request.

### DIFF
--- a/lmcache_ascend/__init__.py
+++ b/lmcache_ascend/__init__.py
@@ -192,6 +192,27 @@ def _patch_config():
         "Set 0 for an unbounded queue; values > 0 enable bounded backpressure.",
     }
 
+    # Add enable_chunk_hashes_return config
+    lmcache.v1.config._CONFIG_DEFINITIONS["enable_chunk_hashes_return"] = {
+        "type": bool,
+        "default": False,
+        "env_converter": _to_bool,
+        "description": "Whether to track chunk hashes during lookup and "
+        "include them in request_finished return params. "
+        "If True, chunk_hashes will be available in return_params. "
+        "Default is False (disabled, no impact on original functionality).",
+    }
+
+    # Add lookup_hashes_cache_size config
+    lmcache.v1.config._CONFIG_DEFINITIONS["lookup_hashes_cache_size"] = {
+        "type": int,
+        "default": 0,
+        "env_converter": int,
+        "description": "Maximum number of cached chunk hash entries. "
+        "When exceeded, the oldest entry is evicted to prevent unbounded "
+        "memory growth. Default is 0 (unlimited).",
+    }
+
     namespace_extras = {
         "validate": lmcache.v1.config._validate_config,
         "log_config": lmcache.v1.config._log_config,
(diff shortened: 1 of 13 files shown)


jianghv:
9b440d77808532cb6f7ae0ad7c366764
include chunk_hashes field in curl request response

matthewygf:
Hmm, I think there might be a better way to re-calculate the hash everytime when `request_finished`.

In theory, 
1. every lookup request already calculate the token hashes per chunk, and this is done at the scheduler side.
2. it is possible to cache the token hashes at the scheduler side all the way till `bind_connector_meta` and attached the chunk_hash then passes to the worker side.
3. At the worker side you can then just get from the request_tracker.

Wdyt ? @jianghv

matthewygf (changes requested):
Left a cmt, on whether we should attached the hash from lookup.

jianghv:
Previous calculations were all done in the worker process, while request_finished is called by the scheduler process, so it cannot be reused.

chloroethylene (changes requested):
Add UT for new feature

jianghv:
402b0a25e038ba6113d78db977fb639b

All the local CI gate checks pass

tests/v1/lookup_client/test_lmcache_ascend_async_lookup_client.py
```diff
@@ -0,0 +1,44 @@
...
+
+class TestLMCacheAscendAsyncLookupClient:
+    @staticmethod
+    def _make_client(**attrs):
+        # First Party
+        from lmcache_ascend.v1.lookup_client.lmcache_ascend_async_lookup_client import (
+            LMCacheAscendAsyncLookupClient,
+        )
+
+        client = LMCacheAscendAsyncLookupClient.__new__(LMCacheAscendAsyncLookupClient)
```
chloroethylene: why not LMCacheAscendAsyncLookupClient() directly construct a object

lmcache_ascend/__init__.py
```diff
@@ -540,6 +530,44 @@ def _patch_rpc_utils():
...
+def _patch_storage_manager():
+    # Third Party
+    from lmcache.v1.storage_backend.storage_manager import StorageManager
+
+    # First Party
+    from lmcache_ascend.v1.storage_backend.storage_manager import (
+        patched_prefetch_all_done_callback,
+    )
+
+    StorageManager.prefetch_all_done_callback = patched_prefetch_all_done_callback
```
chloroethylene, line 565: recommand to change `patched_prefetch_all_done_callback` to `StorageManager_prefetch_all_done_callback`

lmcache_ascend/v1/storage_backend/storage_manager.py
```diff
@@ -0,0 +1,58 @@
+# SPDX-License-Identifier: Apache-2.0
+# Third Party
+from lmcache.logging import init_logger
+from lmcache.v1.event_manager import EventStatus, EventType
+
+logger = init_logger(__name__)
+
+
+def patched_prefetch_all_done_callback(
```
chloroethylene, line 9: add a comment to describe the changed code .

lmcache_ascend/v1/lookup_client/lmcache_ascend_async_lookup_client.py
```diff
@@ -0,0 +1,72 @@
...
+import torch
+
+
+class LMCacheAscendAsyncLookupClient(LMCacheAsyncLookupClient):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # lookup_hashes_cache init start---------------------
+        self._lookup_hashes_cache = {}
+        self._max_cache_size = int(
+            kwargs.pop(
```
chloroethylene: __init__ use the same interface as LMCacheAsyncLookupClient.__init__ . And use `config.xxx` to get config

matthewygf (approved):
LGTM

jianghv:
1.When enable_chunk_hashes_return is True, the prefetch/evict function works correctly.
image

2.When enable_chunk_hashes_return is False, the prefetch/evict interfaces are disabled, which does not affect the original service.
image
3.CI access control verification failed only once, and this failure can pass when run separately
image
image

chloroethylene (approved):

PR #239, opened 2026-05-21
fix: handle kv_events as list in synchronous store mode

Missmiaom:
When store_async is False, self.kv_events remains a plain list (initialized by upstream LMCacheEngine), which does not have a drain() method. Calling drain() unconditionally causes an AttributeError crash on the first inference request.

Use list snapshot-and-clear for synchronous mode and reserve ThreadSafeEventList.drain() for the async path only.


matthewygf (approved):
Thanks for the contribution ! LGTM

PR #243, opened 2026-05-26
fix ci git fetch tiomeout

jianghv:
1. the container needs to enable privileged mode to access the NPU in some environment
2. fix the proxy issue
3. by default, git uses the HTTP/2 protocol. However it may randomly result in issues such as the HTTP/2 layer being lost or connection timeout. switching to HTTP/1.1 resolves these issues


jianghv:
image
it can work normally

matthewygf (approved):
lgtm

PR #244, opened 2026-05-27
fix ci:add git with ssh after checkout in 910c test

jianghv:
ci machine HTTP is unstable, replace git http url with ssh


jianghv:
image

ci is working normally

PR #246, opened 2026-05-28
[Core][SP] Support for multiple and hybrid KV cache groups with mixed type/compression (e.g. Deepseek v4)

marcobarlo:
# [Core] Support for hybrid KV cache groups with mixed type/compression (e.g., Deepseek v4 in vllm-ascend)

## Summary

DeepSeek V4 on vLLM-Ascend ([config](https://docs.vllm.ai/projects/ascend/en/v0.18.0/tutorials/models/DeepSeek-V4-Flash.html)) exposes **11 scheduler KV-cache groups** per request (distinct `block_ids`, `block_size`, `compress_ratio`) and **heterogeneous per-layer tensors** (multi-plane tuples, shared int8 blobs, mixed dtypes).

Upstream [LMCache#3171](https://github.com/LMCache/LMCache/pull/3171) added DSv4 on the **MP** path only. vLLM-Ascend uses the **in-process** connector; upstream `RequestTracker` keeps `block_ids[0]` and one `slot_mapping`, so DSv4 is wrong or fails asserts.

| Feature | Required? | Role |
| ------- | --------- | ---- |
| **1. In-process multi-group support** | Yes | All groups tracked; per-group slots; NPU grouping/dispatch; DSv4 layouts |
| **2. Multi-plane copy bundling** | No (perf) | Fewer NPU launches; default `LMCACHE_ASCEND_BUNDLE_MULTI_SPEC=1` |

---

## Background

| Layer type | `kv_caches[name]` | Scheduler groups |
| ---------- | ----------------- | ---------------- |
| SWA / MTP | 1× `bf16` view | 1 |
| Compress-4 | 7 views, shared `int8` blob | up to 7 |
| Compress-128 | 4-tuple mixed dtypes | up to 4 |

`DS4RandomQuarterLayers`: **11** `KVCacheGroupSpec`s; `new_request.block_ids` is a tuple of 11 lists. Upstream MP (#3171) runs **one CUDA launch per homogeneous group** — it does **not** fuse multiple planes of one layer.

---

## Feature 1 — In-process multi-group support (required)

Without this: groups 1..10 block tables are dropped, compressed `slot_mapping` lengths fail asserts, and `KVLayerGroupsManager` mis-groups blob / mixed-dtype layers. Exploded mode (`LMCACHE_ASCEND_BUNDLE_MULTI_SPEC=0`) is the minimal correct fallback (MP-style, one transfer per homogeneous NPU group).

### Workflow vs upstream

| Step | Upstream in-process | This PR |
| ---- | ------------------- | ------- |
| New request | `allocated_block_ids` = group 0 | `allocated_block_ids_by_group` (all 11) |
| ReqMeta | One `slot_mapping` | `slot_mappings_by_group[g]`; `primary_kv_group_idx` |
| `wait_for_save` / `start_load_kv` | One slot tensor → engine | Per-group CPU/NPU maps + `slot_mappings_npu_by_group` |
| `register_kv_caches` | Flat homogeneous layers | `build_flat_kv_caches` + `layout_hints` |
| Layer groups | `(kv_size, nh, hs, dtype)` | `build_kv_layer_groups` + `KVCacheFormat`; scheduler-slot split |
| Store/load kernels | One pointer table | `_multi_group_kv_transfer`; `scheduler_slot_group` maps NPU index → scheduler group |
| MemoryObj | Single `hidden` / chunk | `_patch_metadata_get_shapes` per NPU group |

MP stages block IDs on the **server**; here the **NPU connector** builds per-group pointers and slot slices.

### End-to-end path

```text
register_kv_caches:   (vllm_v1_adapter) build_flat_kv_caches → layout_hints
                      → ensure_kv_layer_groups (per NPU group MemoryObj sizing)
build_connector_meta: (multi_group_vllm_adapter) RequestTracker + ReqMeta
wait_for_save:        (vllm_v1_adapter) slot_mappings_by_group → NPU → engine.store
                      → connector: multi_layer_kv_transfer[_multi_plane] per NPU group
start_load_kv:        symmetric retrieve (load_stream)
```

Chunk-hash / mask sizing uses the **primary** group’s `slot_mapping`; kernels use **all** groups.

### Main changes

- `lmcache_ascend_connector_v1.py` — entry (`LMCacheAscendConnectorV1Dynamic`, `SupportsHMA`) → `LMCacheAscendConnectorV1Impl`
- `multi_group_vllm_adapter.py` — `LMCacheConnectorV1ImplMultiGroup`, `RequestTracker` / `ReqMeta`, slot builders, `build_connector_meta`
- `vllm_v1_adapter.py` — register, `wait_for_save`, `start_load_kv`, `request_finished*`
- `multi_spec_flatten.py`, `kv_format.py`, `kv_layer_groups.py`, `npu_connectors.py`, `skip_state_groups.py`
- `mem_kernels.cpp` / `utils.cpp` — DSA-C8 sizing on homogeneous kernel path
- `lmcache_ascend/__init__.py` — optional `_patch_vllm_v1_adapter()` shim into upstream `lmcache.integration.vllm.vllm_v1_adapter` on `import lmcache_ascend` (not the DSv4 connector entry path)

### Concepts

- **Plane** — one KV component of a layer (**sub-tensor**: own buffer or shared-allocation view). Bundled layers have several planes; each may map to a different scheduler group.
- **Flat entry** — one slot in LMCache’s per-layer list. **Bundling on:** one tuple per multi-spec layer; **off:** one entry per plane.
- **Same kernel** — iff registered and (a) planes of the **same flat entry**, or (b) different flat entries with the same shape key `(kv_size, hidden, block_size, dtype_key, num_tensors)` and scheduler group.
- **Scheduler KV-cache group** — one vLLM `KVCacheGroupSpec`; `slot_mapping[g]` length `ceil(T / compress_ratio[g])`.
- **NPU layer group** — flat indices sharing shape key **and** scheduler group after split → one kernel per chunk.
- **`scheduler_slot_group`** — NPU group `i` uses `slot_mappings_by_group[g]` where `g` is that group’s scheduler index (not `[i]`).

**DSV4 mapping (bundled, state skipped):**

| Scheduler group | Role | NPU group | Bundled? |
| --------------- | ---- | --------- | -------- |
| sg0, sg1, sg3 | CR4 attn + SWA + indexer (+ scale) | g0 (4 planes) | yes — one flat entry per CR4 layer |
| sg1 @ layer 21 | Standalone SWA | g1 | no |
| sg2 | SWA on even layers ≥22 | g2 | no |
| sg8 | Compress-128 (SW=128) | g3 | no |
| sg4–7, sg10 | State specs | — | skipped at registration |
| sg9 | CR128 SWA companion | — | slot mapping only; not in MemoryObj |

**Example (T=512):** `len(slot_mapping[g]) = ceil(512 / CR[g])`; primary group = `argmax_g(len(block_ids[g]) × block_size[g])`.

### Architectural choices

**Skip state groups** (`skip_state_groups.py`, default on): DSV4 **state** groups (`C4AttnKVState`, `C4IndexerKVState`, `C128AttnScoreState`, …) hold recurrent/score buffers not needed for prefix reuse.

- `LMCACHE_ASCEND_SKIP_STATE_GROUPS=1` — filter before layer-group planning.
- `LMCACHE_ASCEND_SKIP_STATE_SPEC_ALLOWLIST` — optional; default skips six `*State*` specs (sg4–7, sg10).
- **Bundled:** remove skipped planes from tuples; drop empty flat entries.
- **Exploded:** drop flat entries for skipped scheduler groups.
- Indexer KV (`C4Indexer`, sg3) is **not** skipped.

**Sliding-window store** (sg8, SW=128): only the **last** `sliding_window` tokens per LMCache chunk are persisted. Storing the full logical chunk over-allocated MemoryObj rows (~5.2× before fix).

1. **Store mask** (`multi_group_vllm_adapter.py`) — `slot_mapping[g] = -1` for tokens outside `chunk_end - sliding_window`; kernels skip `-1`.
2. **Physical chunk size** (`kv_layer_groups.py`, `npu_connectors.py`, `_patch_metadata_get_shapes`) — `physical_chunk_size = sliding_window // compress_ratio` (SW=128, CR=128 → **1** row for sg8).

Multi-group mode requires `discard_partial_chunks=True` globally.

---

## Feature 2 — Multi-plane copy bundling (optional)

Feature 1 alone is sufficient for correctness via **exploded** mode: each plane is a flat layer; each homogeneous NPU group gets `multi_layer_kv_transfer` — same idea as MP’s multiple CUDA launches. Bundling is a **performance** choice on NPU (launch overhead on Atlas A2), not something upstream MP implements.

| Aspect | Upstream in-process | Upstream MP (#3171) | This PR (bundled) |
| ------ | ------------------- | ------------------- | ----------------- |
| Multi-spec layer | Not supported | N registrations, N launches | One tuple (`MULTI_PLANE_KV` / `DSA_C8_KV`) |
| Heterogeneous `block_size` | N/A | Separate kernels | `multi_layer_kv_transfer_multi_plane` |
| Chunk layout | Standard | Per-group slots | Plane-major `uint8` rows |
| Slot map | Single tensor | Server-built | Concat + `multi_plane_slot_slice_bounds` |

**Files:** `multi_layer_kv_transfer_multi_plane`, `KVCacheFormat.MULTI_PLANE_KV` / `DSA_C8_KV`, `_invoke_multi_plane_kv_transfer`. Launch count: **4 NPU groups** vs exploded O(planes × layers).

---

**Breakage without this PR:**

- `RequestTracker` / `ReqMeta` — single group; wrong retrieve assert.
- `KVLayerGroupsManager` + `gpu_connector/utils.py` — no blob / multi-plane / per-plane `block_size`.
- `VLLMPagedMemGPUConnectorV2` — flat pointer table; mixed tuple lengths break.
- `LMCacheMetadata.get_shapes` — cannot size per-group multi-plane row bytes.
- `multi_layer_kv_transfer` — scalar `block_size` insufficient for bundled planes.
- HMA `request_finished` — needs `request_finished_all_groups` for tuple `block_ids`.

---

## Flags and tests

| Flag | Effect |
| ---- | ------ |
| `LMCACHE_ASCEND_BUNDLE_MULTI_SPEC=0` | Exploded planes; feature 2 off |
| `LMCACHE_ASCEND_FLATTEN_MULTI_SPEC=0` | Legacy single-group; DSv4 unsupported |
| `LMCACHE_ASCEND_SKIP_STATE_GROUPS=1` (default) | Skip state scheduler groups |

**Tests:** `test_ds4_kvcache_roundtrip.py`, `test_multi_group_vllm_adapter.py`, `test_npu_connector_multi_group_load.py`, `test_mem_kernels.py`, `test_kv_layer_groups_npu.py`, `test_per_group_memory_allocation.py`. Single-group models unchanged.

---

### Configuration:  
Official vllm-ascend configuration for DeepseekV4 and in addition:
```
    --no-disable-hybrid-kv-cache-manager \
    --kv-transfer-config '{"kv_connector":"LMCacheAscendConnectorV1Dynamic","kv_role":"kv_both", "kv_connector_module_path":"lmcache_ascend.integration.vllm.lmcache_ascend_connector_v1"}'\
    --no-enable-prefix-caching \
```
LMCache configuration (tag v0.4.5)
```
export LMCACHE_TRACK_USAGE=false
export LMCACHE_MAX_LOCAL_CPU_SIZE="70"
export LMCACHE_LOCAL_CPU="True"
export LMCACHE_LOG_LEVEL=INFO
export LMCACHE_USE_LAYERWISE="False"
export LMCACHE_NUMA_MODE="auto"
export LMCACHE_CHUNK_SIZE=1024
export LMCACHE_EXTRA_CONFIG='{"save_only_first_rank": false}'
```
With the multi group connector defaults to discard partial chunks.

### Results GSM8K
GSM8k results: first run, no LMCache KV cache hit, 
full_GSM8K_FIRST_RUN

GSM8k results: second run, ~ 30 % of requests with KV cache hit form LMCache.
 
full_GSM8K_SECOND_RUN

### Results vllm bench

```
vllm bench serve \
    --host "${HOST}" \
    --port "${PORT}" \
    --backend openai-chat \
    --endpoint /v1/chat/completions \
    --model "${MODEL_NAME}" \
    --tokenizer "${MODEL_PATH}" \
    --tokenizer-mode deepseek_v4 \
    --dataset-name prefix_repetition \
    --num-prompts 100 \
    --request-rate 0.5 \
    --prefix-repetition-num-prefixes 30 \
    --prefix-repetition-prefix-len 30000 \
    --prefix-repetition-suffix-len 10000 \
    --prefix-repetition-output-len 1000 \
    --ignore-eos \
    --temperature 0

```

Results no LMCache (no prefix matching vLLM):
bench_baseline

Results LMCache (no prefix matching vLLM):
bench_lmc

##  EDIT 1 - Upstream HMA ([cec6092](https://github.com/LMCache/LMCache/commit/cec6092db0f034d71df3dd5c16575e1318302e34), [#3419](https://github.com/LMCache/LMCache/pull/3419))

[#3171](https://github.com/LMCache/LMCache/pull/3171) added DSv4 on MP CUDA; **cec6092** generalises to **HMA** (`LMCacheGroupView`, `group_layers_by_identity`, Gemma-3 CI). Same problem class; this PR targets the **in-process NPU** path.

| Concern | Upstream cec6092 | This PR |
| ------- | ---------------- | ------- |
| Block IDs / slots | `expand_block_ids_to_views` | `allocated_block_ids_by_group`, `slot_mappings_by_group[g]` |
| Grouping | `group_layers_by_identity` (CUDA) | `build_kv_layer_groups` + `KVCacheFormat` + scheduler split |
| Physical chunk | `physical_chunk_size` + CR | + `sliding_window_size_by_group` |
| Runtime | MP + CUDA | In-process NPU |
| Layouts / bundling | Homogeneous CUDA tensors | Tuple/blob/multi-plane; `multi_layer_kv_transfer_multi_plane` |
| State / SW | All groups; SW out of scope | Skip `*State*`; tail store + `SW // CR` sizing (~5.2× alloc fix) |

**Net-new:** in-process connector stack (`LMCacheAscendConnectorV1Dynamic` → `LMCacheAscendConnectorV1Impl` → `LMCacheConnectorV1ImplMultiGroup`), Ascend format detection, multi-plane kernel, state skip, SW-aware allocation.

**Integration:** land on LMCache-Ascend for DSv4; on rebase ≥ cec6092 keep in-process adapter + Ascend grouping + skip/SW/bundling (upstream HMA does not replace them). Port engine-neutral SW helpers upstream when DSv4 MP is needed. Cherry-picking cec6092 alone does **not** fix vLLM-Ascend.

## EDIT 2 -- Latest vLLM-Ascend DeepseekV4
vLLM-Ascend has now latest for [DeepseekV4](https://docs.vllm.ai/projects/ascend/en/latest/tutorials/models/DeepSeek-V4-Flash.html). This has likely a different KV cache structure than the v0.18 tested for this PR. Additional tests are needed

## EDIT 3 -- Latest vLLM-AscendDeepseekV4 is now supported (unstable)
The main difference between the two versions are as follows.
[offload_v20_vs_v18_diff.md](https://github.com/user-attachments/files/28892038/offload_v20_vs_v18_diff.md)

Part of the multi-group/multi-plane bundling (feature 2) of this PR may not be needed to support v0.20 and part of it can be stripped out (along with the new kernel). I suggest to merge this PR to a branch of LMCache-Ascend to support DSv4 in v0.18, and then merge only needed parts to continue with upstream from v0.20 on in main branch. The multi-plane path and multi-plane kernel would only be useful to support DSA C8, which is anyway not used in DSv4 (it is used in DSv3.2 and GLM5 with enable_sparse_c8=True).


marcobarlo:
@copilot review

csrc/mem_kernels.cpp
```diff
@@ -143,13 +185,31 @@ void fused_multi_layer_kv_transfer(
...
+
+      kvcache_ops::multi_layer_kv_transfer_multi_plane_kernel_v2(
+          config.aiv_num, config.stream, config.page_buffer_ptrs, staging_cache_ptr,
+          slot_concat_ptr, hidden_dim_bytes.data_ptr<int32_t>(),
+          block_sizes.data_ptr<int32_t>(), page_buffer_sizes.data_ptr<int32_t>(),
+          slot_offsets.data_ptr<int32_t>(), 4, config.num_layers,
+          static_cast<int64_t>(staging_cache.size(-1)) * staging_cache.element_size(),
+          config.num_tokens_lmc_chunk, config.singlePerLoopBuffer,
+          config.maxTokensPerLoop, config.direction);
+      ret = aclrtSynchronizeStream(config.stream);
```
matthewygf: do we need to sync ?

marcobarlo: No, should be fixed with new commit. WIP testing

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1,293 +1,617 @@
...
+
+            if request.load_spec is None or not request.load_spec.can_load:
+                continue
+
+            tokens = request.token_ids
+            slot_mappings_cpu: list[torch.Tensor] = []
+            for group_idx in range(request.num_kv_groups):
+                group_slot_mapping = request.get_slot_mapping(group_idx)
+                assert isinstance(group_slot_mapping, torch.Tensor)
+                slot_mappings_cpu.append(group_slot_mapping.pin_memory())
```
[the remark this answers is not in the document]
marcobarlo: @matthewygf I remember you played with it. Can we keep blocking copy or should be move to single-pinned-storage shared across tensors with async copy?

lmcache_ascend/v1/memory_management.py
matthewygf, line 1: @chloroethylene @marcobarlo are we sure this should be deleted. This seems like needed for the slang patch ? Double checking if we need it for sgl.

lmcache_ascend/integration/vllm/lmcache_ascend_connector_v1.py
```diff
@@ -18,10 +24,40 @@
...
+# First Party
+from lmcache_ascend.integration.vllm.vllm_v1_adapter import LMCacheAscendConnectorV1Impl
 
 logger = init_logger(__name__)
 
 
-class LMCacheAscendConnectorV1Dynamic(LMCacheConnectorV1Dynamic):
-    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole) -> None:
-        super().__init__(vllm_config=vllm_config, role=role)
+class LMCacheAscendConnectorV1Dynamic(LMCacheConnectorV1Dynamic, SupportsHMA):
```
matthewygf, line 35: We will need to add 'SupportsHMA' to the entry connector from vllm-ascend.
Two options:
1. make a PR in vllm-ascend for the in-process connector to have SupportsHMA
2. create a connector in LMCache-Ascend, patched it during the import lmcache-ascend.

wdyt @marcobarlo @chloroethylene 
It's probably better to go with 1.

matthewygf: as discussed, let's go with 2) since in process mode is going to be deprecated.

matthewygf (changes requested):

docs/deployment.md
```diff
@@ -132,4 +126,114 @@ ulimit -l unlimited # Update with the amount of memory you need to lock in KBs
...
+    -v /etc/ascend_install.info:/etc/ascend_install.info \
+    -v /usr/local/Ascend/driver:/usr/local/Ascend/driver \
+    -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
+    -v /lib/modules:/lib/modules:ro \
+    -v /usr/src/kernels:/usr/src/kernels:ro \
+    -v /mnt/storage1/data:/data \
+    -v /home:/home \
+    --name lmcache-ascend-ms \
+    --entrypoint /bin/bash \
+    lmcache-ascend:v0.4.4-mindspore2.7.1.post1-openeuler
```
matthewygf, line 168: As discussed, change this back to v0.3.12 for actual tested ver.

docs/deployment.md
```diff
@@ -132,4 +126,114 @@ ulimit -l unlimited # Update with the amount of memory you need to lock in KBs
...
+#### Online serving
+```bash
+python \
+    -m vllm_mindspore.entrypoints vllm.entrypoints.openai.api_server \
+    --port 8100 \
+    --model /data/models/Qwen/Qwen3-32B \
+    --trust-remote-code \
+    --disable-log-requests \
+    --block-size 128 \
+    --kv-transfer-config '{"kv_connector":"LMCacheAscendConnectorV1Dynamic","kv_role":"kv_both", "kv_connector_module_path":"lmcache_ascend.integration.vllm.lmcache_ascend_connector_v1"}'
```
matthewygf: change this to use the in process connector in vllm-ascend

lmcache_ascend/v1/slot_mapping_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+    For each scheduler group, strips ``-1`` (invalid) rows from ``sm`` and
+    builds a prefix array where ``prefix[g][k]`` counts valid slots in
+    ``sm[g][0:k)``. Enables O(1) per-chunk dense slicing via
+    :func:`dense_bounds_from_prefix` instead of scanning ``-1`` on every
+    transfer.
+    """
+    ratios = tuple(int(r) for r in compress_ratios)
+    if not ratios:
+        ratios = (1,) * len(slot_mappings_by_group)
```
matthewygf: unused ?

matthewygf (changes requested):

lmcache_ascend/integration/vllm/multi_spec_flatten.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+def flatten_multi_spec_enabled() -> bool:
+    val = os.environ.get("LMCACHE_ASCEND_FLATTEN_MULTI_SPEC", "1")
+    return val != "0"
+
+
+def bundle_multi_spec_enabled() -> bool:
+    val = os.environ.get("LMCACHE_ASCEND_BUNDLE_MULTI_SPEC", "1")
+    return val != "0"
```
matthewygf: we should probably change these into the config definition as per __init__.py 

like:

```
lmcache.v1.config._CONFIG_DEFINITIONS["ascend_bundle_multi_spec"] = {
    "type": bool,
    "default": True,
    "env_converter": _to_bool,
    "description": (
        "Whether LMCache-Ascend keeps multi-spec KV planes bundled for "
        "multi-plane NPU transfer. If False, multi-spec planes are exploded "
        "into synthetic .subN layers for legacy/fallback handling."
    ),
}
```

matthewygf (changes requested):

lmcache_ascend/integration/vllm/skip_state_groups.py
```diff
@@ -0,0 +1,283 @@
...
+    suffix = getattr(policy, "layer_name_suffix", None)
+    if suffix is not None:
+        return suffix
+    return DEFAULT_SKIP_STATE_SUFFIX
+
+
+def parse_skip_state_policy_from_env() -> SkipStateGroupsPolicy | None:
+    """Read skip policy from env and return None when the feature is disabled."""
+    if os.environ.get("LMCACHE_ASCEND_SKIP_STATE_GROUPS", "1") != "1":
+        return None
```
matthewygf: Similarly, It would be best if we can make these configurable in the config definition __init__.py

matthewygf (changes requested):

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
```diff
@@ -22,32 +28,283 @@
...
+                # NOTE(Jiayi): Perform blending before layerwise prefix caching
+                if self.enable_blending:
+                    self.blender.blend(
+                        tokens[:lmcache_cached_tokens],
+                        token_mask[:lmcache_cached_tokens],
+                        kvcaches=kvcaches,
+                        slot_mapping=slot_mapping_npu[:lmcache_cached_tokens],
+                        vllm_cached_tokens=request.load_spec.vllm_cached_tokens,
+                    )
+                else:
```
matthewygf: Not sure whether we should enable blending here tbh for kvcache groups > 1

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+        if klg_manager is None or not klg_manager.kv_layer_groups:
+            self.group_kv_cache_pointers = None
+            self.per_group_params = None
+            return
+
+        group_pointers: list[torch.Tensor] = []
+        group_params: list[dict[str, Any]] = []
+        for group_idx, group in enumerate(klg_manager.kv_layer_groups):
+            indices = group.layer_indices
+            rep = kv_caches[indices[0]]
```
matthewygf, line 1078: Is this always true to be indices[0] ? Might worth to have a small comment for this assumption

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -30,6 +40,308 @@
...
+            bs_list.append(bs)
+            pbs_list.append(slots)
+            # Per-token byte width of this plane: total plane bytes / page slots.
+            pb.append(int(t.numel() * t.element_size()) // slots if slots else 0)
+    else:
+        pb = [int(x) for x in plane_hidden_bytes or ()]
+    n = len(pb)
+    if kv_format == KVCacheFormat.DSA_C8_KV and n != 4:
+        raise ValueError(f"DSA-C8 expects 4 plane byte widths, got {n}")
+    if not bs_list:
```
matthewygf: nit: might be `Len(bs_list) == 0` or `planes is not None` ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -30,6 +40,308 @@
...
+            else:
+                raise ValueError(
+                    f"Unsupported SEPARATE_KV entry for multi-plane: {type(entry)}"
+                )
+            if is_310p():
+                block_size = int(k_cache.shape[-2])
+                page_buffer_size = int(k_cache.shape[0]) * block_size
+            else:
+                block_size = int(k_cache.shape[1])
+                page_buffer_size = int(k_cache.shape[0]) * block_size
```
matthewygf: nit: trivial, maybe worth taking the page_buffer_size out of the if branches

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -30,6 +40,308 @@
...
+            page_buffer_size=page_buffer_size,
+        )
+    elif entry_format == KVCacheFormat.MLA_KV:
+        sched_groups = (layout_hints or {}).get("layer_to_scheduler_groups", {})
+        layer_name = (layout_hints or {}).get("_current_layer_name")
+        sched_per_plane: list[int] = []
+        if layer_name and layer_name in sched_groups:
+            sched_per_plane = list(sched_groups[layer_name])
+            if len(sched_per_plane) == 1:
+                sched_per_plane = sched_per_plane * 2
```
matthewygf: is it the kv lora and rope ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -452,6 +782,16 @@ def __init__(
             self.dtype = kwargs["dtype"]
             self.device = kwargs["device"]
 
+    def initialize_kvcaches_ptr(self, **kwargs) -> None:
+        """Initialize KV cache pointers using Ascend-safe permute for pool slices."""
+        if "kvcaches" in kwargs:
+            self.kvcaches = kwargs["kvcaches"]
+            from lmcache_ascend.v1.npu_connector.utils import (
+                permute_kv_caches_to_contiguous,
+            )
```
matthewygf: does this have to be lazily import actually...

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+        if self.metadata is None or self.metadata.kv_layer_groups_manager is not None:
+            self._sync_logical_page_slots_from_manager()
+            return
+
+        # Third Party
+        from lmcache.v1.gpu_connector.utils import normalize_kv_and_discover_format
+        from lmcache.v1.kv_layer_groups import KVLayerGroupsManager
+        from lmcache.utils import EngineType
+
+        from lmcache_ascend.v1.kv_layer_groups import build_kv_layer_groups
```
matthewygf: Also not need to be lazily imported I think

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+        if getattr(self, "_mp_launch_bufs", None) is None:
+            per_group_params = getattr(self, "per_group_params", None)
+            n = (
+                len(per_group_params)
+                if per_group_params
+                else npu_group_idx + 1
+            )
+            self._mp_launch_bufs = [None] * n
+        while len(self._mp_launch_bufs) <= npu_group_idx:
+            self._mp_launch_bufs.append(None)
```
matthewygf: nit: its better to switch to a compact form, maybe like 

```
missing = npu_group_idx + 1 - Len(self._mp_launch_bufs)
if missing> 0:
  self._mp_launch_bufs.extend([None]*missing)
```

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+            mem_tensor = mem_tensor.view(torch.uint8)
+        num_planes = group_params["num_planes"]
+
+        device = filtered_slot_mappings_npu[0].device
+        bufs = self._ensure_mp_launch_bufs(npu_group_idx, num_planes, device)
+        key = (g_start, g_end, npu_group_idx)
+        if mp_launch_meta is not None:
+            cached = mp_launch_meta.get(key)
+            if cached is None:
+                return
```
matthewygf, line 1004: This is very unclear, it seems it meant it is known there is no work for the cached key from build_mp_launch_meta ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+            starts_cpu, counts_cpu, has_work = compute_mp_plane_launch_row(
+                g_start,
+                g_end,
+                sched_groups,
+                slot_mappings_by_group=slot_mappings_by_group,
+                prefixes_by_group=slot_valid_prefix_by_group,
+                compress_ratios=compress_ratios,
+            )
+            if not has_work:
+                return
```
matthewygf, line 1022: probably worth a logger ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -491,8 +831,341 @@ def from_metadata(
...
+                        entry_format,
+                        multi_plane=multi_plane_ptrs,
+                    )
+                )
+            cpu_ptrs = torch.empty(len(ptrs), dtype=torch.int64, device="cpu")
+            cpu_ptrs.numpy()[:] = ptrs
+            gpu_ptrs = torch.empty(
+                len(ptrs), dtype=torch.int64, device=self.kvcaches_device
+            )
+            gpu_ptrs.copy_(cpu_ptrs)
```
matthewygf, line 1103: worth making these async ? were there any perf hits ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -605,8 +1339,38 @@ def _initialize_pointers(self, kv_caches: List[torch.Tensor]) -> torch.Tensor:
...
+        entry = caches[0]
+        k_cache = entry[0]
+        return _derive_group_params(
+            entry,
+            KVCacheFormat.DSA_C8_KV,
+            SimpleNamespace(
+                nb=int(k_cache.shape[0]),
+                bs=int(k_cache.shape[1]),
+                block_stride_elems=0,
+            ),
```
matthewygf: nit: while it is okay, not a big fan of simplenamespace, maybe even refactor the function to explicitly take nb, bs, block_stride is better

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -791,6 +1565,196 @@ def to_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+        filtered = kwargs.get("filtered_slot_mappings_npu")
+        prefixes = kwargs.get("slot_valid_prefix_by_group")
+        slot_mappings_by_group = kwargs.get("slot_mappings_by_group")
+        if (
+            filtered is None
+            or prefixes is None
+            or slot_mappings_by_group is None
+            or not starts
+        ):
+            return
```
matthewygf, line 1610: should we raise an error here? not clear to me when do these conditions should be true or not......

matthewygf (changes requested):

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -1091,23 +2093,21 @@ def batched_from_gpu(self, memory_objs, starts, ends, **kwargs):
...
-        end_event.record(self.store_stream)
         self.store_stream.synchronize()
-        from_gpu_time = start_event.elapsed_time(end_event)
-
-        return from_gpu_time / 1000.0  # convert to seconds
 
     def get_shape(self, num_tokens: int) -> torch.Size:
+        if self.kv_format == KVCacheFormat.DSA_C8_KV:
+            from lmcache_ascend.v1.kv_layer_groups import _lmc_chunk_hidden_bytes
+
```
matthewygf: let's move this to the module level import as well.

lmcache_ascend/v1/kv_format.py
```diff
@@ -1,14 +1,106 @@
...
+
+# Extracts the block_size dimension from a KV plane tensor (dim-1 for 3-D/4-D).
+# Used by _is_multi_plane_tuple and the NPU kernel param builder to derive
+# per-plane paging geometry.
+def _plane_block_size(tensor: torch.Tensor) -> int:
+    # block_size lives at dim-1 for both 3-D (nb, bs, hidden) and
+    # 4-D (nb, bs, nh, hs) layouts; ndim < 3 has no paging geometry.
+    if tensor.ndim >= 3:
+        return int(tensor.shape[1])
+    raise ValueError(f"Unexpected KV plane ndim={tensor.ndim}")
```
matthewygf: nit: if we can somehow pass in a tensor format hint somewhere, this will be better , i.e. the vLLM hints etc wdyt @marcobarlo

lmcache_ascend/v1/kv_layer_groups.py
```diff
@@ -1,140 +1,387 @@
...
+    vllm_block_size: int,
+    *,
+    is_310p: bool = False,
+) -> torch.Size:
+    """Derive ``[num_blocks, block_size, hidden]`` from a shared-storage blob.
+
+    Multiple dtype views alias one allocation; only the primary view (largest byte
+    coverage) carries the canonical paging geometry.
+    """
+    del vllm_block_size
```
matthewygf: what is this `del vllm_block_size` ??

lmcache_ascend/v1/kv_layer_groups.py
```diff
@@ -1,140 +1,387 @@
...
+    hidden_per_token = int(primary.numel()) // (num_blocks * block_size)
+    return torch.Size([num_blocks, block_size, hidden_per_token])
+
+
+def _get_kv_cache_group_key_and_info(
+    kv_cache: _LayerKV,
+    *,
+    is_310p: bool = False,
+    vllm_block_size: Optional[int] = None,
+) -> tuple[int, int, int, Any, int]:
```
matthewygf: nit: the tuple[int, int, int, Any, int] is really hard to understand :/ , probably better to have a named type

lmcache_ascend/v1/kv_layer_groups.py
```diff
@@ -1,140 +1,387 @@
...
+            if sched_map is not None
+            else shape_key
+        )
+        groups_dict[key].append(idx)
+
+    # Sort groups by the first layer index to maintain order
+    def _get_first_layer_index(key):
+        return groups_dict[key][0]
+
+    sorted_keys = sorted(groups_dict.keys(), key=_get_first_layer_index)
```
matthewygf: maybe:

```
# Sort groups by first layer index to maintain deterministic flat-kv order.
sorted_keys = sorted(groups_dict, key=lambda key: groups_dict[key][0])
```

matthewygf (changes requested):

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
```diff
@@ -0,0 +1,787 @@
...
+    }
+
+    // Copy one contiguous byte span from UB at srcByteOff into Paged GM via DataCopyPad.
+    // No branches; used as the MTE3 consumer step on the load path inside the depth-2 pipeline.
+    __aicore__ inline void CopyUbToPaged(AscendC::GlobalTensor<uint8_t> &dst,
+        const int64_t dstByteOff, AscendC::LocalTensor<uint8_t> &src,
+        const int64_t srcByteOff, const uint32_t copyBytes) const {
+        const AscendC::DataCopyExtParams params{1u, copyBytes, 0u, 0u, 0u};
+        AscendC::DataCopyPad(dst[dstByteOff], src[srcByteOff], params);
+    }
```
matthewygf, line 178: as discussed, these two probably can be merged... since the diff seems to be the PadExtParams. wdyt

lmcache_ascend/v1/memory_management.py
marcobarlo: Related to https://github.com/LMCache/LMCache-Ascend/pull/246/changes/531c8c18517283fa66f06ed84824bb3118404118#diff-fd20e68221eb641e878f120ac8c9115639848b45d9477476008f6f672da8fb9c @chloroethylene please check

lmcache_ascend/v1/slot_mapping_utils.py
marcobarlo: Leftover from refactoring. compress_ratios function parameter is also unused. To be removed

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
marcobarlo: That code path is there since it was copy-pasted from super.start_load_kv() and extended for multi-group, but it is actually untested - and probably we do not want to test it as SP is going to be deprecated. Flagging for deletion.

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: Flagging to add the following comment: 
```
            # ``build_kv_layer_groups`` buckets layers by layout identity
            # (kv_size, hidden, block_size, dtype, tensor count). Every member
            # of ``indices`` shares that key and ``group.shape_desc``, so format
            # detection and ``_derive_group_params`` on ``indices[0]`` apply to
            # the whole group; per-layer device pointers are collected below.
```
Does it clarify?

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: preferrable ```len(bs_list) == 0``` as ```planes is not None``` is already the gating condition of the data structure building above.

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
```diff
@@ -0,0 +1,787 @@
...
+        const uint32_t partBytesU32 =
+            static_cast<uint32_t>(static_cast<int64_t>(partNumTokens) * hdBytes);
+        CopyUbToPaged(pagedGm, pagedByteStart, scratchU8, 0, partBytesU32);
+        this->pagedTokenQue_.FreeTensor(buf);
+    }
+
+    // Copy one paged-block part on the blockwise (hd<32) path using the depth-2 queue when bulk fits UB.
+    // Branch bulk: store EnQue Paged->UB then flush prev to Lmc, or load EnQue Lmc->UB then flush prev to
+    // Paged; branch !bulk: drain pending then copyBlockSetValue (no UB pipeline for this part).
+    __aicore__ inline void copyBlock(
```
matthewygf: The copyBlock function works, but it would be better for readability wise since Page2L and L2Page is merged into single function and the if branches exit via `return`.

Perhaps it is better to separate into two helpers like 
`copyBlockPageToLmc` and `copyBlockLmcToPage` ?

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: No. MLA_KV format is generically detected any time the KV cache is a tuple with 2 tensors and the two tensors have different shape (```k_cache.shape != v_cache.shape```). 
In the standard MLA path, this translates to lora and rope, but in that case we use the standard copy path and we do not enter the multi-group path. Since this snippet is in _derive_group_params, it means we are in heterogeneous/multi-group path: each plane is a generically a buffer of bytes with its own geometry. As example, in v0.20 the indexer is a tuple of int8 in K with last dim 128, and FP 16 scale with last dim 1 (copying the indexer might not be strictly necessary tho...).

Probably a cleaner version of the snippet would be:
```
n_planes = len(entry)
if len(sched_per_plane) == 1 and n_planes > 1:
    sched_per_plane = sched_per_plane * n_planes
```
but at the moment there is a little difference because to detect MLA n_planes must be 2. 
@matthewygf  wdyt?

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
```diff
@@ -0,0 +1,787 @@
...
+            } else {
+                flushPendingUbPartToPaged(
+                    pagedGm, pendingPagedByteStart, pendingNumTokens, hdBytes);
+            }
+        }
+    }
+
+    // Store one UB window: per block-part, pipeline Paged->UB (EnQue) with UB->Lmc (DeQue prev).
+    // Branch bulk: EnQue after CopyPagedToUb; branch invalid lead: drain pending then SetValue only.
+    __aicore__ inline void _page2LTransfer(__gm__ uint8_t *cacheTensor, __gm__ uint8_t *slotmappings,
```
matthewygf, line 543: nit: the inner function code could be merged I think between _page2L and L2Page, maybe via a templated class. could be a separate PR for later opt

matthewygf (changes requested):

README.md
matthewygf, line 1: Probably should remove ref to the Dynamic Connector once we made support in the in process one.

matthewygf (changes requested):

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: No need, we anyway need to import at least once. Probably leftover from refactoring.  Flag to move above.

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: Likewise. Flag to move them as module-level imports.

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: This line is quite of a safety net: all NPU groups are iterated for each memObj. However, some groups might have no work to do (e.g., all tokens stripped out or compressed or other borderline cases). build_mp_launch_meta in this cases inserts no key in meta, and at invocation time we skip the kernel call. Probably worth adding a comment 

```
            # Batch precompute omits keys with no valid slots (has_work=False);
            # same as mp_launch_meta is None path — skip the kernel for this chunk/group.
```
and add a log level debug in such case (see next comment).

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: See comment above

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: It is a one off cold-path on the first copy. Probably not worth the hustle

lmcache_ascend/v1/npu_connector/npu_connectors.py
marcobarlo: No. This is just a shortcut for normal non-multi-group transfers that use the normal non-multi-plane copy kernel to skip this as not needed. Furthermore, this function is just an optimization. In case the meta is needed by the multi-plane kernel, it is created just in time.

lmcache_ascend/v1/kv_format.py
marcobarlo: The concept of cache plane lives only in LMCache, vLLM only exposes block_sizes_by_group (vLLM scheduler group), which may or may not differ from LMCache NPU groups. For this reason, it is non-trivial to pass through such a hint. The most straightforward way is to keep it like this and make it more robust (e.g., is_310p formats and so on).

lmcache_ascend/v1/kv_layer_groups.py
marcobarlo: Leftover from refactor. Flag to remove and remove from function signature.

lmcache_ascend/v1/kv_layer_groups.py
marcobarlo: Agreed. Flag to use named tuple.

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
marcobarlo: We could, but probably would hurt readability because we need a direction flag as parameter instead of having a self-explaining function name. Probably the best is to simply remove the two helpers at all and directly inline DataCopyExtParams -> DataCopyPad

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
marcobarlo: what about blockwiseCopy? Shuld we separate that as well? Wdyt? Probably best to separate only copyBlock.

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
marcobarlo: At the moment is mirroring the structure of the multi_layer_mem_kernels_v2.cpp. Probably best to take a common decision for both kernel so it is easier to follow them.

marcobarlo:
Rebased on main and tested with gsm8k for vLLM v0.20.

lmcache_ascend/v1/npu_connector/npu_connectors.py
matthewygf: The cleaner version looks a lot better tbh !

new_kernels/multi_layer_mem_kernels_v2_multi_plane.cpp
matthewygf: Yeah agreed. BlockWiseCopy seems fine to me tbh.

marcobarlo:
Closing PR as it was superseded by https://github.com/LMCache/LMCache-Ascend/pull/271 that forked this branch. The support is now merged in https://github.com/LMCache/LMCache-Ascend/tree/dsv4_support_045

PR #254, opened 2026-06-16
Add hole-mode CacheBlend extension for non-contiguous cache reuse

scestola-h:
## Summary

Adds hole-mode CacheBlend to LMCache-Ascend: extends the existing CacheBlend path to support reuse of cached KV across non-contiguous prefixes (with "holes"). Standard CacheBlend stops at the first cache miss and recomputes everything after it. Hole-mode computes fresh KV for the missing segment(s) only and continues to reuse cached segments that follow.

## Where to start reading

The maintainer overview for this PR is at [`docs/hole-feature-overview.md`](docs/hole-feature-overview.md). It covers the architecture, request flow vs standard CacheBlend, file map, and a recommended code-review reading order.

## What's included

- Hole-aware connector and adapter (`LMCacheConnectorV1ImplHole`)
- Segment-aware blender (`LMCBlenderHole`)
- Hole-aware NPU buffer connector (`VLLMBufferLayerwiseNPUHoleConnector`)
- Hole-aware lookup client and server
- Supporting types and segment utilities under `lmcache_ascend/v1/`
- Benchmark entrypoint (`benchmark/cacheblend/benchmark-hole.py`) with config (`benchmark/cacheblend/config/config-hole.yaml`)
- Test (`tests/v1/blend/test_blend.py`)
- Maintainer overview (`docs/hole-feature-overview.md`)

## Design

The standard CacheBlend path is preserved unchanged. Hole-mode is selected via connector module choice. See `docs/hole-feature-overview.md` for design rationale and where each piece of the request flow lives.

## Validation

Single-hole and multi-hole runs on the Musique multi-hop QA dataset with Qwen3-8B. On a matched comparison (inflate=20, until=150, recompute_ratio=0.15), standard CacheBlend reaches f1=0.2230; single-hole mode reaches f1=0.2200. Concrete numbers and configuration details in section 6 of the overview doc.


chloroethylene:
@larksudo review plz

scestola-h:
# Hole-mode CacheBlend: Analysis and Reproduction

Following up on the request for precision, coverage, and speed data on this PR. All numbers below are from runs on Ascend 910B4 NPU hardware using the benchmark harness included in this PR (`benchmark/cacheblend/benchmark-hole.py`). The model used across every run below is Qwen3-8B, and the dataset is Musique.

If this data is insufficient for review, we're happy to run additional experiments. For example, broader parameter sweeps, additional recompute ratios or hole configurations, other datasets, or additional models. Let us know what would be most useful.

## 1. What the feature does (recap)

Standard LMCache CacheBlend reuses a contiguous cached KV prefix and stops at the first cache miss. Hole-mode extends this to reuse cached KV across non-contiguous prefixes: when a prompt has one or more "holes" (missing cached document positions) in an otherwise cached prefix, hole-mode computes fresh KV for the holes and continues to reuse cached segments that follow.

## 2. Definitions

### 2.1 Dataset and prompt structure

**Musique**: multi-hop QA dataset. Each query is a question paired with a set of native retrieved documents. The model is asked to answer the question given the documents. The benchmark serialises this as:

    <system> # # <doc_1> # # <doc_2> # # ... # # <question>

### 2.2 Configuration parameters

- **`inflate` (N)**: appends N additional documents to each query's native document set. These extra documents come from a shared pool of non-native documents. `inflate=1` keeps native docs unchanged; `inflate=10` adds 10 extra documents; `inflate=30` adds 30. This tests long-context behavior without changing dataset semantics.
- **`hole_id`**: index into the query's native retrieved documents, selecting which one becomes the "hole document."
- **`hole_pos`**: position in the final inflated document sequence where the hole document is placed. Its KV cache is not populated during phase 1, which forces hole-mode (or, if hole-mode is disabled, forces a legacy full-recompute) at inference time.
- **`hole_ids`, `hole_positions`**: multi-hole variant. E.g., `hole_ids=[2,5] hole_positions=[3,6]` means native doc 2 is placed at inflated position 3 and native doc 5 is placed at inflated position 6, both with unpopulated KV.

Concrete walkthrough: with `hole_id=2`, `hole_pos=3`, `inflate=10`, the benchmark takes the query's native retrieved docs, appends 10 extra pool docs, and constructs the final inflated list. Native doc 2 is placed at position 3 in that list. That single position's KV is not populated during phase 1. At inference, the model must either activate hole-mode (recompute KV for position 3 while reusing the rest) or fall back to the legacy full-recompute path.
- **`hole_mode`**: `hole` enables the hole-mode path; `no-hole` runs baseline CacheBlend.
- **`recompute_ratio` (r)**: standard CacheBlend parameter. Fraction of reusable tokens recomputed at the check layer. Lower r = faster inference but less quality-corrective recompute. Default in shipped config: 0.5. Tighter regimes tested here: 0.3, 0.2, 0.15.
- **`until` (N)**: number of queries drawn from Musique to run in each benchmark invocation. All numbers below use `until=150`.
- **`check layer`**: single layer at which CacheBlend performs its diff/topK check to decide which cached-KV positions need refresh. Configurable via `blend_check_layers`; the shipped benchmark config uses layer 1.

### 2.3 Metrics

Each benchmark run does two phases:

- **Phase 1**: batched prefill of documents into the LMCache cache.
- **Phase 2**: query processing. Two passes over the query set:
  - **Pass 1 (`phase2_pass1`)**: first full traversal; fills any missing hole chunks during processing.
  - **Pass 2 (`phase2_pass2`)**: second full traversal; measures steady-state reuse behavior with warmed cache.

Metrics reported per pass:

- **`mean_f1`**: mean F1 over the answer set, using Musique's evaluation. Higher is better.
- **`cache_hit_rate`**: fraction of prompt tokens served from cached KV, computed as `exported_hit_tokens / exported_prompt_tokens`.
- **`latency_avg_e2e_s`**: average end-to-end latency per request in seconds.

*Note on pass 2 `cache_hit_rate`*: pass 2 is not 100% by design. The benchmark deliberately modifies the query suffix between passes (appends `?` per pass) to prevent trivial query-side cache reuse. The reported ~99.7–99.9% figure reflects reuse of the retrieved-document prefix, which is what CacheBlend actually targets.

### 2.4 Modes

- **`method=cacheblend`**: LMCache-backed inference with `LMCACHE_ENABLE_BLENDING=True`. This is the CacheBlend path: batched prefill + blend-based reuse.
- **`method=reuse`**: LMCache-backed inference with `LMCACHE_ENABLE_BLENDING=False`. Same prefill infrastructure but no blending, pure cache reuse without CacheBlend's recompute step.
- **`method=full`**: plain vLLM with no LMCache. Serves as a quality ceiling reference (no reuse artifacts).

## 3. F1 (precision) analysis

### 3.1 Short context, r=0.5 — hole positions and multi-hole

`inflate=1` (no extra documents added). "Baseline" here means LMCache CacheBlend with `hole_mode=no-hole`, i.e., the standard contiguous-prefix CacheBlend path that this PR extends.

| Configuration | hole_mode | Hole config | r | f1 p1 | f1 p2 | hit p1 | hit p2 | e2e p1 (s) | e2e p2 (s) |
|---|---|---|---:|---:|---:|---:|---:|---:|---:|
| row1 (baseline) | no-hole | — | 0.5 | 0.3523 | 0.3657 | 0.6038 | 0.9962 | 0.9467 | 0.7721 |
| row2 (hole @ pos 3) | hole | hole_id=2, hole_pos=3 | 0.5 | 0.3388 | 0.3499 | 0.9261 | 0.9979 | 0.9313 | 0.8293 |
| row3 (hole @ pos 6) | hole | hole_id=2, hole_pos=6 | 0.5 | 0.3289 | 0.3369 | 0.9261 | 0.9979 | 0.9314 | 0.7726 |
| row4 (hole @ pos 8) | hole | hole_id=2, hole_pos=8 | 0.5 | 0.3369 | 0.3372 | 0.9261 | 0.9979 | 0.9851 | 0.7900 |
| row5 (two holes) | hole | hole_ids=[2,5], hole_positions=[4,7] | 0.5 | 0.3448 | 0.3422 | 0.8582 | 0.9991 | 1.0187 | 0.7778 |

Observations:

- Baseline f1 p2: 0.3657. Single-hole variants: 0.3369–0.3499. Multi-hole: 0.3422.
- All hole configurations land within 0.03 of the baseline; multi-hole within 0.024.


### 3.2 Recompute-ratio sensitivity (r=0.3)

Same short-context regime, tighter recompute ratio.

| Configuration | hole_mode | r | f1 p1 | f1 p2 | hit p1 | hit p2 | e2e p1 (s) | e2e p2 (s) |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| row6 (baseline) | no-hole | 0.3 | 0.3397 | 0.3228 | 0.6038 | 0.9962 | 0.9368 | 0.7152 |
| row7 (hole @ pos 6) | hole | 0.3 | 0.3039 | 0.3109 | 0.9261 | 0.9979 | 0.8992 | 0.8222 |

Observation: at r=0.3, baseline f1 drops to 0.3228 (from 0.3657 at r=0.5), the expected quality/speed tradeoff. Hole runs remain within ~0.012 of the baseline.

### 3.3 Long context (inflate=10, 20, 30) at r=0.15

The most stress-testing regime. Inflate synthesizes long contexts by adding extra documents; r=0.15 is a tight-recompute setting. As in section 3.1, "baseline" here means LMCache CacheBlend with `hole_mode=no-hole`.

| Configuration | hole_mode | inflate | r | f1 p1 | f1 p2 | hit p1 | hit p2 | e2e p1 (s) | e2e p2 (s) |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| row16 (baseline) | no-hole | 10 | 0.15 | 0.2587 | 0.2378 | 0.6359 | 0.9972 | 1.2258 | 0.8712 |
| row17 (hole @ pos 4) | hole | 10 | 0.15 | 0.2227 | 0.2222 | 0.9587 | 0.9988 | 1.0103 | 0.8457 |
| row18 (baseline) | no-hole | 20 | 0.15 | 0.2263 | 0.2230 | 0.6405 | 0.9975 | 1.7630 | 1.1434 |
| row19 (hole @ pos 6) | hole | 20 | 0.15 | 0.2167 | 0.2148 | 0.9717 | 0.9992 | 1.4411 | 1.1566 |
| row23 (hole @ pos 4) | hole | 20 | 0.15 | 0.2230 | 0.2200 | 0.9717 | 0.9992 | 1.3748 | 1.1385 |
| row21 (baseline) | no-hole | 30 | 0.15 | 0.2023 | 0.1916 | 0.6422 | 0.9977 | 2.4660 | 1.5244 |
| row22 (hole @ pos 4) | hole | 30 | 0.15 | 0.1758 | 0.1753 | 0.9784 | 0.9996 | 1.8335 | 1.5139 |

Observations:

- Absolute f1 decreases with more inflated context (harder task), this happens for baseline and hole alike.
- The baseline-vs-hole f1 gap remains within ~0.05 across the entire long-context regime.


### 3.4 Reference points (full vLLM, LMCache-only)

Quality ceiling (full vLLM, no cache reuse) and cache-reuse-without-blending baseline.

| Configuration | method | inflate | r | f1 p1 | f1 p2 | hit p1 | hit p2 | e2e p1 (s) | e2e p2 (s) |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| row8 | full | 1 | 0.5 | 0.3760 | 0.3760 | 0.0000 | 0.0000 | 0.8600 | 0.8557 |
| row9 | reuse (no CacheBlend) | 1 | 0.5 | 0.3760 | 0.3807 | 0.0009 | 0.9982 | 0.9772 | 0.5922 |
| row10 | full | 10 | 0.5 | 0.3783 | 0.3783 | 0.0000 | 0.0000 | 1.4588 | 1.4627 |
| row13 | full | 20 | 0.5 | 0.3252 | 0.3252 | 0.0000 | 0.0000 | 2.3049 | 2.3106 |
| row20 | full | 30 | 0.5 | 0.3046 | 0.3046 | 0.0000 | 0.0000 | 3.3118 | 3.3167 |

Observations:

- Full vLLM at inflate=1: f1=0.3760, latency 0.8557s. Baseline CacheBlend (row1) reaches f1=0.3657, latency 0.7721s, trading ~0.01 f1 for ~10% speedup.
- Reuse-without-blending (row9) at f1=0.3807 slightly exceeds full vLLM (0.3760) at inflate=1, with much lower latency. The small f1 difference is within measurement noise for a 150-query set. This mode uses CacheBlend's cache-reuse infrastructure without the blend/recompute step.

## 4. Hole path activation (coverage evidence)

Two dedicated validation runs (`val_A_single_hole` and `val_B_two_holes`) were configured to match a single-hole and multi-hole scenario respectively. val_A processes ~21K requests and val_B processes ~19K requests through Phase 2.

- `val_A_single_hole` configuration: `hole_mode=hole hole_id=2 hole_pos=4 inflate=20 recompute_ratio=0.15 until=150`. Same configuration as row23.
- `val_B_two_holes` configuration: `hole_mode=hole hole_ids=[2,5] hole_positions=[3,6] inflate=20 recompute_ratio=0.15 until=150`.

### 4.1 Per-request path distribution

Each request that reaches the LMCache connector is routed to one of three paths depending on its cache state:

| Path | val_A (single hole) | val_B (two holes) | What it means |
|---|---:|---:|---|
| Hole path (log: `Hole mode=hole`) | 108 | 130 | Hole-mode logic activated; fresh KV computed for hole positions, cached KV reused elsewhere. |
| Legacy path (log: `Hole mode=legacy`) | 1148 | 1035 | The hole-mode adapter routed the request through the legacy contiguous-prefix path. |
| Pure-hit path (log: `load_mode=pure_hit`) | 20559 | 18325 | Full LMCache reuse without needing hole-mode; all documents were cached. |

### 4.2 Coverage statistics on hole-path requests

For requests that took the hole path:

| Metric | val_A | val_B |
|---|---|---|
| Covered tokens (range) | 0 – 18,995 | 0 – 19,002 |
| Prefix-miss tokens (range) | 0 – 879 | 0 – 1,697 |

These show that when hole-mode engages, it covers up to ~19K tokens of the reusable prefix, with the hole itself accounting for at most ~1.7K tokens of fresh recompute.

### 4.3 Quality of validation runs

| Configuration | f1 p1 | f1 p2 | hit p2 | e2e p2 (s) |
|---|---:|---:|---:|---:|
| val_A (single hole) | 0.2230 | 0.2200 | 0.9992 | 1.1089 |
| val_B (two holes) | 0.2022 | 0.2022 | 0.9996 | 1.1281 |

val_A reproduces row23's quality (f1 p2 = 0.2200) and cache behavior (hit p2 = 0.9992); latency differs by ~0.03s, within typical run-to-run variance. val_B provides a matched multi-hole datapoint at the same context/r configuration.

## 5. Reproduction

### 5.1 Environment

- Hardware: Ascend 910B4 NPU (single card, NPU 3 used for all matrix runs)
- npu-smi driver version: 25.6.rc1.b099
- Feature branch: this PR at commit 58b0133
- Container base image: `quay.io/ascend/vllm-ascend:v0.18.0`
- Python package versions:
  - vllm 0.18.0+empty
  - vllm-ascend 0.18.0
  - lmcache 0.4.4
  - torch 2.9.0+cpu
  - torch_npu 2.9.0.post1+gitee7ba04

### 5.2 Model and dataset

- Model: Qwen3-8B 
- Dataset: Musique multi-hop QA 

### 5.3 Benchmark entrypoint

Included in this PR:

    benchmark/cacheblend/benchmark-hole.py
    benchmark/cacheblend/config/config-hole.yaml

### 5.4 Row-by-row commands

Each row is reproduced by:

    python3 benchmark/cacheblend/benchmark-hole.py <override>

Where `<override>` is:

| Row | Override |
|---|---|
| row1 (baseline, inflate=1) | `npu=3 method=cacheblend hole_mode=no-hole inflate=1 until=150` |
| row2 (hole @ pos 3) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=3 inflate=1 until=150` |
| row3 (hole @ pos 6) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=6 inflate=1 until=150` |
| row4 (hole @ pos 8) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=8 inflate=1 until=150` |
| row5 (two holes) | `npu=3 method=cacheblend hole_mode=hole hole_ids=[2,5] hole_positions=[4,7] inflate=1 until=150` |
| row6 (baseline, r=0.3) | `npu=3 method=cacheblend hole_mode=no-hole inflate=1 until=150 recompute_ratio=0.3` |
| row7 (hole, r=0.3) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=6 inflate=1 until=150 recompute_ratio=0.3` |
| row8 (full vLLM) | `npu=3 method=full hole_mode=no-hole inflate=1 until=150` |
| row9 (reuse) | `npu=3 method=reuse hole_mode=no-hole inflate=1 until=150` |
| row10 (full, inflate=10) | `npu=3 method=full hole_mode=no-hole inflate=10 until=150` |
| row13 (full, inflate=20) | `npu=3 method=full hole_mode=no-hole inflate=20 until=150` |
| row16 (baseline, inflate=10, r=0.15) | `npu=3 method=cacheblend hole_mode=no-hole inflate=10 recompute_ratio=0.15 until=150` |
| row17 (hole, inflate=10, r=0.15) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=4 inflate=10 recompute_ratio=0.15 until=150` |
| row18 (baseline, inflate=20, r=0.15) | `npu=3 method=cacheblend hole_mode=no-hole inflate=20 recompute_ratio=0.15 until=150` |
| row19 (hole, inflate=20, r=0.15) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=6 inflate=20 recompute_ratio=0.15 until=150` |
| row20 (full, inflate=30) | `npu=3 method=full hole_mode=no-hole inflate=30 until=150` |
| row21 (baseline, inflate=30, r=0.15) | `npu=3 method=cacheblend hole_mode=no-hole inflate=30 recompute_ratio=0.15 until=150` |
| row22 (hole, inflate=30, r=0.15) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=4 inflate=30 recompute_ratio=0.15 until=150` |
| row23 (hole, inflate=20, r=0.15) | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=4 inflate=20 recompute_ratio=0.15 until=150` |
| val_A_single_hole | `npu=3 method=cacheblend hole_mode=hole hole_id=2 hole_pos=4 inflate=20 recompute_ratio=0.15 until=150` |
| val_B_two_holes | `npu=3 method=cacheblend hole_mode=hole hole_ids=[2,5] hole_positions=[3,6] inflate=20 recompute_ratio=0.15 until=150` |

Notes:

- All runs use `until=150` (150 queries drawn from Musique).
- `hole_mode=no-hole` selects the standard contiguous-prefix CacheBlend path.
- `hole_mode=hole` selects the hole-mode path introduced in this PR.
- Baseline recompute_ratio (from shipped config) is 0.5.

benchmark/v1/rag/README.md
```diff
@@ -16,9 +16,9 @@ python /LMCache-Ascend/lmcache_ascend/integration/patch/apply_patch.py
 #### Option 2: Manual Modification
 If you prefer to update the code manually, please modify the following file:
 
-**File Path**: `vllm-ascend/vllm-ascend/worker/worker_v1.py`
+**File Path**: `vllm-ascend/vllm-ascend/worker/worker.py`
 
-- In `vllm-ascend/vllm-ascend/worker/worker_v1.py`, comment out `ensure_kv_transfer_initialized(vllm_config)` in function `def _init_worker_distributed_environment`.
+- In `vllm-ascend/vllm-ascend/worker/worker.py`, comment out `ensure_kv_transfer_initialized(self.vllm_config, kv_cache_config)` in function `def initialize_from_config`.
```
larksudo, line 23: Nice work. I see the changes on the latest main branch.
For completeness, should we list which vllm-ascend versions are supported.

scestola-h: Good point, will add a version compatibility note. Addressing this along with a few other pending items asap.

lmcache_ascend/v1/cache_engine.py
```diff
@@ -425,6 +426,137 @@ def get_kv_events(self) -> Iterable[CacheStoreEvent]:
...
+            for layer_id in range(self.num_layers):
+                yield None
+
+        yield None
+
+        next(mem_obj_consumer)
+
+        for mem_obj in to_count_down:
+            if mem_obj.is_pinned:
+                mem_obj.unpin()
```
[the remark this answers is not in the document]
scestola-h: Fixed in 221138e , wrapped the finalization block in `if keys:` so the `next(mem_obj_consumer)` and `to_count_down` iteration only run on the retrieval path where those variables are actually defined.

lmcache_ascend/v1/blend/blender.py
```diff
@@ -112,6 +182,22 @@ def process_qkv(
...
+            trace_flow(
+                "blender.nohole",
+                "process_qkv",
+                layer_id=layer_id,
+                req_id=kwargs.get("req_id", self._trace_req_id),
+                num_falses=num_falses,
+                eligible_tokens=total_len,
+                topk_num=topk_num,
+                top_indices=top_indices.tolist(),
+                absolute_top_indices=absolute_top_indices.tolist(),
```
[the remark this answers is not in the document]
scestola-h: This one appears to be a false positive on the current branch. In `process_qkv`, `total_len` is assigned at line 177 (`total_len = diff_k.shape[0]`) inside the same `if (layer_id in self.common_metadata.check_layers and self.common_metadata.recomp_ratios[0] > 0):` block that contains its uses on lines 189 and 196 so there is no reachable path in the function that reaches those uses without `total_len` being assigned.

Note: `main` has since refactored this section (introducing `_get_recomp_ratio(layer_id)`). Once the branch is rebased onto `main`, the code structure will change and this comment may need to be re-evaluated against the merged code.

lmcache_ascend/__init__.py
```diff
@@ -695,12 +815,27 @@ def _patch_api_server():
...
+        import lmcache.v1.trace_utils  # noqa: F401
+    except ModuleNotFoundError as exc:
+        if exc.name != "lmcache.v1.trace_utils":
+            raise
+
+        # First Party
+        from lmcache_ascend.v1 import trace_utils_compat
+
+        sys.modules["lmcache.v1.trace_utils"] = trace_utils_compat
+
```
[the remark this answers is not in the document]
scestola-h: Addressed in 221138e. The patch now covers both cases: when `lmcache.v1.trace_utils` is missing entirely (the original behavior, install our compat module as a stand-in), and when it exists but lacks newer symbols (backfill any missing attributes from our compat module by iterating over `dir(_compat)` and skipping private names). This avoids hardcoding a symbol list and picks up any new helpers we add later automatically.

benchmark/cacheblend/benchmark-hole.py
```diff
@@ -0,0 +1,1658 @@
...
+        timer_rows.append(timer_metrics)
+        output_str = output.outputs[0].text
+        answers = meta['answers']
+
+        e2e_time = extract_request_e2e_s(output)
+        ttft = extract_request_ttft_s(output)
+        tbt = None
+        if e2e_time is not None and ttft is not None:
+            generated_tokens = len(output.outputs[0].token_ids)
+            tbt = (e2e_time - ttft) / max(1, generated_tokens - 1)
```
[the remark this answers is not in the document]
scestola-h: Fixed in 221138e , added a guard that logs a warning and skips the request via `continue` if `output.outputs` is empty. Slightly more conservative than the suggested fallback-to-empty-string approach , avoids emitting a zero-length answer into the evaluation set.

lmcache_ascend/utils.py
```diff
@@ -0,0 +1,224 @@
...
+        return None  # Not found
+    except subprocess.CalledProcessError as e:
+        print("Error running npu-smi:", e)
+        return None
+
+def get_mem(npu_id, chip_id=0):
+    try:
+        return parse_npu_process_info(npu_id,chip_id)['memory_mb']
+    except:
+        return -1
```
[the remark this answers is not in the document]
scestola-h: Fixed in 221138e. Changed to `except Exception:` per PEP 8.

scestola-h:
All 6 review comments addressed in 221138e:
- 4 code fixes (cache_engine, __init__, benchmark-hole, utils)
- 1 documentation note (README compatibility)
- 1 false-positive explanation (blender.py:199)

Verified with smoke tests on Ascend 910B4: row1 (standard CacheBlend) and row23 (hole mode, inflate=20, r=0.15) both reproduce previously recorded baselines to 3 decimals on F1 and hit rate.

Two follow-ups still pending, will address separately:
- Merge conflicts on blender.py and cache_engine.py (main added `_get_recomp_ratio` and a sharded-broadcast refactor)
- Code Quality CI (pre-existing lint / SPDX / isort issues across multiple files, unrelated to this batch)

scestola-h:
Merged `main` into this branch (merge commit 51588b5).

Two files required manual conflict resolution:
- **`blender.py`**: kept main's #253 changes (`_get_recomp_ratio` + validation) while preserving this PR's timer instrumentation and the `gpu_connector_override=self.gpu_connector` argument in `blend_layer`'s `retrieve_layer` call, required for the hole connector's legacy fallback path.
- **`cache_engine.py`**: kept main's #257 sharded broadcast refactor of `retrieve()` while preserving `AscendLMCacheEngine.retrieve_layer`. That override honors `gpu_connector_override` (upstream ignores it) and is needed for the hole legacy fallback per its docstring. Retained the empty-key cache-miss guard from 221138e.

Smoke tests on Ascend 910B4 reproduce baseline F1:
- row1 (standard, `inflate=1`, `r=0.5`): 0.365672 vs baseline 0.3657
- row23 (hole, `inflate=20`, `hole_pos=4`, `r=0.15`): 0.219973 vs baseline 0.2200

scestola-h:
@larksudo   flagging you here in case you have some bandwidth to take a look. I've addressed the initial comments and merged in latest `main`. If you get the chance to share thoughts on the overall design of the hole-mode implementation, that would be really helpful.

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
```diff
@@ -25,9 +29,77 @@
...
+) -> Optional[dict[str, Any]]:
+    if not base_params and not extra_params:
+        return None
+    merged: dict[str, Any] = {}
+    if base_params:
+        merged.update(base_params)
+    if extra_params:
+        merged.update(extra_params)
+    return merged
+
```
larksudo: The metrics utility functions (`_set_request_output_metrics`, `_collect_request_output_metrics`, `_merge_request_output_params`) are added to the shared `vllm_v1_adapter.py` but only used by the hole connector. The standard path now carries `HoleSegmentHelper` as a dependency for metrics it doesn't need. Either move these helpers to a separate utils module, or keep them inside `vllm_v1_adapter_hole.py` to keep the standard path untouched.

This applies to all shared files touched by this PR — please audit every modification to non-hole files (e.g. `blender.py`, `cache_engine.py`, `positional_encoding.py`, `models/*.py`) and apply the same principle: if the change only benefits the hole path, it should live in hole-specific code, not in shared infrastructure.

lmcache_ascend/integration/vllm/vllm_v1_adapter_hole.py
```diff
@@ -0,0 +1,1509 @@
...
+from vllm.config import VllmConfig
+from vllm.distributed.kv_transfer.kv_connector.v1.base import (
+    KVConnectorBase_V1,
+    KVConnectorRole,
+)
+from vllm.distributed.parallel_state import (
+    get_pp_group,
+    get_tensor_model_parallel_rank,
+    get_tp_group,
+)
```
larksudo: Run pre-commit --all-files before pushing. This will catch code quality issues like unused imports (e.g. get_pp_group , get_tensor_model_parallel_rank ), formatting, and linting violations automatically.

lmcache_ascend/v1/lookup_client/lmcache_hole_lookup_server.py
```diff
@@ -0,0 +1,269 @@
...
+                    hit_flags=hit_flags,
+                    mode=result.mode,
+                    covered_tokens=result.covered_tokens,
+                    hit_ranges=summarize_ranges(result.hit_ranges),
+                    prefix_miss_ranges=summarize_ranges(result.prefix_miss_ranges),
+                    speculative_tail_segments=len(segment_ranges) - 1,
+                    speculative_fallback=False,
+                )
+                return result
+
```
larksudo, line 216: The _lookup_tokens method has 4 early-return branches, each duplicating the same derive_lookup_result + trace_flow + return block. Consider extracting a single _finish_lookup() helper and calling it from each branch — this would cut the method roughly in half and make the branching logic much easier to follow.

lmcache_ascend/v1/lookup_client/lmcache_hole_lookup_server.py
```diff
@@ -0,0 +1,269 @@
...
+                covered_tokens=result.covered_tokens,
+                hit_ranges=summarize_ranges(result.hit_ranges),
+                prefix_miss_ranges=summarize_ranges(result.prefix_miss_ranges),
+                speculative_tail_segments=0,
+                speculative_fallback=False,
+            )
+            return result
+
+        if first_miss_idx is not None:
+            if first_miss_idx == 0:
```
larksudo, line 203: When segment 0 is a miss, _lookup_tokens immediately returns legacy mode without checking remaining segments. Could you explain the reasoning behind this? Is it purely to avoid unnecessary storage lookups, or is there a correctness concern?

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
scestola-h: Addressed in 54f7404. I moved `_merge_request_output_params`, `_set_request_output_metrics`, `_collect_request_output_metrics`, and the `HoleSegmentHelper` dependency out of the shared `vllm_v1_adapter.py` and into `vllm_v1_adapter_hole.py`. The standard adapter and standard connector now match `origin/main` exactly. I also audited the remaining shared-file changes; those retained are integration hooks required for connector selection/model dispatch or for the legacy fallback path, rather than hole-only metrics implementation. The merged result passes the complete pre-commit suite plus standard and hole-path NPU smoke tests.

lmcache_ascend/integration/vllm/vllm_v1_adapter_hole.py
scestola-h: Addressed in 54f7404. I removed the unused imports and related unused local state, then ran `PRE_COMMIT_HOME=/tmp/pre-commit-cache pre-commit run --all-files`. SPDX, isort, Ruff, Ruff format, codespell, clang-format, and mypy all pass.

lmcache_ascend/v1/lookup_client/lmcache_hole_lookup_server.py
scestola-h: Addressed in 54f7404. `_lookup_tokens` now delegates the repeated `derive_lookup_result` + `trace_flow` + return sequence to a single `_finish_lookup()` helper, and every early-return branch uses it. I also added focused regression coverage for first-segment and later-segment misses.

lmcache_ascend/v1/lookup_client/lmcache_hole_lookup_server.py
scestola-h: This is a correctness constraint, not only a storage-lookup optimization. Hole mode must expose a materialized covered prefix anchored at token 0 to the scheduler. If segment 0 is missing, hits in later segments cannot be represented as a usable cached prefix, so probing or pinning those segments would produce work that the request cannot consume. The correct fallback is therefore legacy mode immediately. Commit 54f7404 documents this invariant at the branch and adds a test that verifies a first-segment miss performs only one storage probe; a companion test verifies that a later miss can still produce hole mode when the initial covered prefix exists.

benchmark/v1/rag/README.md
scestola-h: Added and retained the compatibility note at the top of `benchmark/v1/rag/README.md`: this feature is developed and tested against `vllm-ascend v0.18.0`. The note is present in the current PR head 54f7404.

scestola-h:
Updated in 54f7404 and merged the current `origin/main` (through 169ee72). All latest review feedback is addressed and the corresponding threads have detailed replies.

Verification completed before push:
- `pre-commit run --all-files`: SPDX, isort, Ruff, Ruff format, codespell, clang-format, and mypy all pass.
- Standard CacheBlend smoke test (row1): f1=0.365672, hit=0.997882, e2e=0.78098s.
- Hole-mode smoke test (row23): f1=0.219973, hit=0.999199, e2e=1.14614s.
- GitHub reports the PR mergeable with no branch conflicts.

lmcache_ascend/v1/blend/hole_blender.py
```diff
@@ -0,0 +1,1363 @@
...
+            gap_trace,
+            top_hit_trace,
+            recompute_trace,
+            token_ids[:50],
+        )
+        logger.info(
+            "#recomp_ids=%d #hit_positions=%d #gap_positions=%d",
+            int(recompute_positions.numel()),
+            int(hit_positions.numel()),
+            int(gap_positions.numel()),
```
larksudo, line 505: Logging is overly verbose across hole-related files. Many logger.info() calls in hole_blender.py and vllm_v1_adapter_hole.py emit per-layer, per-request diagnostic detail (e.g., shape info, gap/hit counts, recomputed token IDs) that should be logger.debug() or removed entirely. In production batch serving, these would flood logs. Additionally, timing/trace instrumentation is scattered across 6+ files with 3 inconsistent APIs (emit_blend_timer, emit_blend_component, emit_request_timer). Consider unifying into a single abstraction and downgrading diagnostic logging to DEBUG level？

lmcache_ascend/v1/cache_engine.py
```diff
@@ -725,6 +736,138 @@ def retrieve(
             )
         return ret_mask
 
+    @torch.inference_mode()
+    def retrieve_layer(
+        self,
```
larksudo: Several shared files (cache_engine.py, blender.py, vllm_v1_adapter.py) are modified solely to support hole-mode's legacy fallback path (e.g., retrieve_layer override for gpu_connector_override, timer state in blender.py). Since hole-mode already has its own dedicated files, consider moving these overrides into hole-specific subclasses rather than modifying shared code — this keeps the standard path completely untouched when hole mode is disabled.

lmcache_ascend/v1/npu_connector/npu_connectors.py
```diff
@@ -56,6 +75,179 @@ def __init__(
         self.kv_format: KVCacheFormat = KVCacheFormat.UNDEFINED
         self.use_mla = bool(kwargs.get("use_mla", False))
         self.fused_rotary_emb: Any = None
+        self._fine_grained_sync_enabled = _env_enabled(
+            "LMCACHE_HOLE_FINE_GRAINED_SYNC",
+            default=True,
+        )
+        self._fine_grained_sync_runtime_disabled = False
```
larksudo: Since npu_hole_connector.py already provides a dedicated hole-mode connector, consider moving the CacheBlend-specific connector logic that hole-mode depends on (currently imported from npu_connector.py) into the hole file itself. This keeps npu_connector.py on the main branch clean and maintainable, and avoids coupling the standard path to hole-mode requirements.

lmcache_ascend/v1/blend/positional_encoding.py
```diff
@@ -119,6 +119,31 @@ def __call__(self, old_positions, new_positions, k):
         return self.fused_encode(old_positions, new_positions, k)
 
 
+def validate_fused_correctness(rope, fused_rope, head_size: int) -> bool:
```
larksudo: validate_fused_correctness appears to be a subset of the existing validate_reverse_correctness

scestola-h:
Thanks for the additional feedback. I’m addressing the new review comments now, including isolating hole-specific behavior from the shared engine/connector paths and consolidating the timer and validation logic. I’ll push an updated revision once the changes and smoke tests are complete.

lmcache_ascend/v1/blend/hole_blender.py
scestola-h: Addressed in 9d197a5. I consolidated hole-mode timing behind a single `lmcache_ascend.v1.timer.emit_timer` API, which dispatches request- and layer-level timers consistently. Hole blender, connector, and adapter callsites now use this abstraction; the previous `emit_blend_timer` / `emit_blend_component` feature callsites are gone. Per-request and per-layer diagnostics in `hole_blender.py` and `vllm_v1_adapter_hole.py` were removed or downgraded to DEBUG/gated trace output; the remaining INFO messages are one-time initialization summaries. Verification: full pre-commit passes, including Ruff and mypy, and the focused pytest suite reports 7 passed.

lmcache_ascend/v1/cache_engine.py
scestola-h: Addressed in 9d197a5. The standard `cache_engine.py`, `blender.py`, `vllm_v1_adapter.py`, `npu_connector/npu_connectors.py`, and standard blend tests now match `origin/main` exactly. The legacy layerwise retrieval override moved to the hole-only `HoleLegacyCacheEngine`; hole-specific attention completion moved to `blend/models/hole.py`; and `vllm_v1_adapter_hole.py` wires those adapters only when hole mode is active. The remaining shared model/RoPE differences are generic vLLM-Ascend 0.18 compatibility used by both standard and hole blending, not hole-only callbacks. Standard Qwen and Llama CacheBlend paths were smoke-tested after this isolation.

lmcache_ascend/v1/npu_connector/npu_connectors.py
scestola-h: Addressed in 9d197a5. `lmcache_ascend/v1/npu_connector/npu_connectors.py` now matches `origin/main` exactly. The legacy fallback connector behavior was moved into `VLLMBufferLayerwiseNPULegacyHoleConnector` in `npu_hole_connector.py`, alongside the sparse hole connector and its hole-specific position-offset storage and stream/event synchronization. `vllm_v1_adapter_hole.py` imports and constructs both connectors from the hole module, so the standard connector no longer carries hole-mode requirements.

lmcache_ascend/v1/blend/positional_encoding.py
scestola-h: Addressed in 9d197a5. I extracted the common fused relocation check into `_validate_fused_relocation`. Both `validate_fused_correctness` and `validate_reverse_correctness` now call that helper, so the former no longer duplicates the fused subset of the latter. The live-RoPE validation path was exercised successfully by Qwen3-8B and Llama-3.1-8B standard and hole-mode NPU smoke tests.

scestola-h:
Follow-up revision pushed in 9d197a5. This addresses the four latest review threads:

- restored the standard cache engine, blender, vLLM adapter, NPU connector, and standard blend tests to `origin/main`; legacy retrieval and synchronization now live in hole-specific adapters
- moved the Qwen/Llama attention-completion hook into hole-only model wrappers
- consolidated request/layer timing behind one `emit_timer` API and reduced production log verbosity
- shared the fused RoPE relocation validation instead of duplicating it

Verification on the final code:
- `pre-commit run --all-files`: all checks pass, including Ruff and mypy
- focused pytest suite: 7 passed
- Qwen3-8B full row23: f1=0.219973, hit=0.999199 (matches baseline)
- Llama-3.1-8B standard and hole-mode smoke tests both completed; hole pass 1 exercised non-contiguous reuse and pass 2 reached hit=1.0

Each review thread has a specific implementation reply and is now resolved.

lmcache_ascend/v1/blend/hole_blender.py
```diff
@@ -0,0 +1,1362 @@
...
+        )
+        plan_lookup_start = time.perf_counter()
+
+        if self.cache_engine.storage_manager is None:
+            raise ValueError("storage_manager is required for hole retrieve")
+
+        for start, end in effective_hit_ranges:
+            if end <= prefix_start:
+                continue
+            adj_start = max(start, prefix_start)
```
matthewygf, line 766: **High — a vLLM prefix ending inside a hit segment changes the cache key and turns the lookup hit into a load miss.**

The hole lookup hashes each complete semantic range `(start, end)`, but here `prefix_start = load_spec.vllm_cached_tokens` changes that range to `(prefix_start, end)` whenever `start < prefix_start < end`. `make_cache_key()` therefore hashes different token bytes from the key that lookup confirmed and pinned. `layerwise_batched_get()` will normally return `None` for that suffix key, and `batched_to_gpu()` then dereferences the missing object; a coincidental suffix entry would not make this a reliable contract.

For example, with hit range `[0, 500)` and `vllm_cached_tokens=128`, lookup confirms `hash(tokens[0:500])`, while retrieval asks for `hash(tokens[128:500])`.

Could we either (1) retain the original full-segment key and carry a source offset so only `[128:500)` is copied into the dense buffer, or (2) explicitly reject/disable nonzero vLLM prefix reuse for hole mode? Please add a regression where `prefix_start` falls strictly inside a hit range. The current benchmark sets `enable_prefix_caching=False`, so its NPU runs do not cover this branch even though the adapter accepts and propagates arbitrary `num_computed_tokens`.

matthewygf (commented):
I found one correctness gap in the hole-only retrieval plan when vLLM already has a nonzero cached prefix. The inline comment contains the concrete failure sequence and suggested regression coverage.

lmcache_ascend/v1/blend/hole_blender.py
scestola-h: Addressed in bc128b2. Hole retrieval now keeps the cache key anchored to the original semantic hit range and carries a per-range source offset to the hole connector. When a vLLM prefix ends inside a hit segment, the connector copies only the requested suffix of K, V, and cached-position metadata into the adjusted dense destination range; ranges not intersected by the prefix retain offset 0 behavior.

I added the concrete regression from this thread: hit range `[0, 500)` with `prefix_start=128`. It asserts that retrieval still requests the `[0, 500)` key, passes destination `[128, 500)` with source offset `128`, and slices cached tokens `[128:500)`. Verification: the focused hole suite passes (`9 passed`) and `pre-commit run --all-files` passes all hooks. The benchmark still has prefix caching disabled, so the new regression test is the branch-specific coverage for this case.

scestola-h:
@larksudo Gentle follow-up when you have time. Thanks again for your patience while I was away. Since returning, I’ve addressed the remaining feedback, resolved all review threads, and verified the latest revision with focused tests and pre-commit. The Code Quality workflow is currently waiting for repository approval. Please let me know if anything else is needed from my side.

PR #257, opened 2026-06-26
Sharded Broadcast with Dual-Stream Pipeline for Reduced Peak NPU Memory

larksudo:
## Summary

This PR introduces **sharded broadcast** for KV cache retrieval in multi-rank (TP) deployments. Instead of broadcasting all KV cache chunks in a single collective and then transferring them to GPU, the broadcast is split into smaller shards that are pipelined through two NPU streams — `broadcast_stream` and `load_stream` — enabling overlap between communication and GPU-side data transfer.

## Motivation

The upstream implementation broadcasts all KV cache chunks at once, which causes two problems:

1. **Peak NPU memory pressure**: all receive buffers for the entire KV cache are allocated simultaneously, driving up peak memory usage linearly with context length.
2. **No overlap**: broadcast and `to_gpu` execute sequentially — the GPU is idle during broadcast, and the NPU is idle during data transfer.

This PR addresses both by sharding the broadcast and pipelining it with `to_gpu` on separate streams.

### `cache_engine.py`

- Added `_pipelined_sharded_broadcast_and_load()` as the core entry point for sharded broadcast.
- `_pipeline_sender()` (rank 0) and `_pipeline_receiver()` (non-rank-0) handle the per-shard loop.
- A dedicated `broadcast_stream` is created when `save_only_first_rank=True` and a `load_stream` is available on the GPU connector.

### Dual-stream pipeline

Each shard records an `Event` on `broadcast_stream` after the broadcast completes. The `load_stream` waits on this event before executing `to_gpu`. The CPU dispatcher immediately moves to the next shard after submitting `to_gpu`, allowing the NPU to overlap the next shard's broadcast with the current shard's `to_gpu`.

```
Time →

broadcast_stream:  |--bc shard[0]--|--bc shard[1]--|--bc shard[2]--|...
load_stream:                        |--to_gpu shard[0]--|--to_gpu shard[1]--|...
```

### Per-shard memory allocation

NPU receive buffers are allocated per shard rather than for the entire KV cache at once. A `_try_release_pending()` mechanism reclaims each shard's buffers once its `to_gpu` event fires, keeping peak NPU memory bounded at approximately `2 × shard_size × chunk_size`.

### Rank 0 H2D optimization

CPU-to-NPU data transfer on rank 0 uses `to(non_blocking=True)` with pinned memory, avoiding synchronous H2D copies that previously dominated broadcast latency.

## Configuration

Add the following to your LMCache config YAML:

```yaml
chunk_size: 256
local_cpu: True
max_local_cpu_size: 2

extra_config:
  save_only_first_rank: True
  first_rank_max_local_cpu_size: 50
  broadcast_shard_size: 8   # number of chunks per broadcast shard (default: 16)
```

### Parameters

| Parameter | Description | Default |
|---|---|---|
| `save_only_first_rank` | Only rank 0 persists KV cache to storage. Enables the dual-stream pipeline; non-rank-0 workers skip store and lookup operations, reducing CPU-side overhead. | `False` |
| `broadcast_shard_size` | Number of chunks per broadcast shard. Smaller values reduce peak NPU memory but increase the number of shard iterations. Set to `0` or a value larger than the total chunk count to fall back to the original single-broadcast behavior. | `16` |

## Trade-offs

| `broadcast_shard_size` | Peak NPU Memory | Pipeline Overhead |
|---|---|---|
| Small (e.g., 4) | Lower | More shard iterations |
| Large (e.g., 32) | Higher | Fewer shard iterations |
| 0 / max | Same as upstream | No sharding (original behavior) |

A value of `8` or `16` is recommended for most workloads.


lmcache_ascend/v1/cache_engine.py
```diff
@@ -133,6 +139,507 @@ def post_init(self, **kwargs) -> None:
...
+
+        def _submit_togpu(ctx):
+            """Enqueue ctx's to_gpu on load_stream and release its slot."""
+            objs, starts, ends, ev_bc, idx, slot = ctx
+            load_stream.wait_event(ev_bc)
+            with torch.npu.stream(load_stream):
+                for obj, s, e in zip(objs, starts, ends, strict=False):
+                    self.gpu_connector.to_gpu(obj, s, e, **kwargs)
+                ev_togpu = torch.npu.Event()
+                ev_togpu.record()
```
matthewygf: does this has to be an inline function ?

lmcache_ascend/v1/cache_engine.py
```diff
@@ -133,6 +139,507 @@ def post_init(self, **kwargs) -> None:
...
+            if prev_ctx is not None:
+                _submit_togpu(prev_ctx)
+
+        finally:
+            # All to_gpu events in pending depend (transitively) on the
+            # H2D copies on broadcast_stream, so once these synchronize
+            # the CPU mem_objs are safe to release.
+            for objs, ev in pending:
+                try:
+                    ev.synchronize()
```
matthewygf: While I think this is safe, I wonder whether we should save some overhead of looping through the pending and call ev.synchronize() or we just should call the load_stream.synchronize() since load_stream sync should be the last execution of the pipeline line ? wdyt ?

matthewygf (changes requested):

matthewygf (approved):
lgtm

Issue #263, opened 2026-07-13
[BUG] Crash due to race condition in HCCL peer handshake


marcobarlo:
**Describe the bug**

HCCL peer handshake in `HcclChannel` races between the ZMQ init loop and the background `accept()` thread. `HcclMemRegRequest` can be handled before `conn_handles_dict[peer_id]` is published, so the decoder returns `HcclErrorResponse` and the prefiller fails with `AttributeError: 'HcclErrorResponse' object has no attribute 'server_mem_handle_bytes'`. PD KV transfer then fails and the disagg proxy hangs waiting for KV-ready.

**To Reproduce**

1. Launch Ascend PD disagg (prefiller `kv_producer` + decoder `kv_consumer`, `transfer_channel: hccl`).
2. Send a chat/completions request through the LMCache disagg proxy (e.g. `tests/query3.sh`).
3. Prefiller calls `lazy_init_peer_connection` → Init → `connect()` → MemReg.
4. See decoder `Failed to process hccl initialization loop: '<peer_id>'` and prefiller MemReg decode error.

**Expected behavior**

MemReg must run only after `hccl_agent.accept()` has stored the peer handle (same ordering as `HcclReadyRequest` / `ready_event`), then exchange buffer handles and complete PD transfer.

**Platform & Software (please complete the following information):**
 - OS: Ubuntu 22.04.5 LTS (container, aarch64)
 - CANN Version: 9.0.0
 - Ascend HDK Version: [fill from local Ascend install]
 - Hardware: Ascend 910B (TP8 prefiller + TP8 decoder)
 - Framework: PyTorch / vLLM-Ascend
 - vLLM Version: ~0.20.2 (Ascend build)
 - LMCache / LMCache-Ascend: v0.4.5-aligned Ascend plugin
 - Model: DeepSeek-V4-Flash (MLA)

**Additional context**
 - Root cause: On `HcclInitRequest`, `_handle_init_msg` starts `complete_handshake()` (background `accept()`), waits only for `accept_started_event`, then replies Init. Client may finish `connect()` and send MemReg before T3 writes `conn_handles_dict`.
 - Fix: In `HcclMemRegRequest` handling, `ready_event.wait()` before looking up `conn_handles_dict`.
 - logs:
```
decoder: Processing HcclMemRegRequest
decoder: Failed to process hccl initialization loop: 'localhostXXXX'
decoder: Background: Connection established with localhostXXXX
[...]

PR #266, opened 2026-07-13
[Bug] Fix possible race condition in HCCL handshake

marcobarlo:
## Summary

This PR fixes issue https://github.com/LMCache/LMCache-Ascend/issues/263.

- Fix a race in Ascend HCCL PD peer handshake: `HcclMemRegRequest` can run
  before the background `accept()` thread publishes `conn_handles_dict[peer_id]`.
- Decoder then returns `HcclErrorResponse`; prefiller crashes with
  `AttributeError: 'HcclErrorResponse' object has no attribute
  'server_mem_handle_bytes'`. PD transfer fails; proxy hangs on KV-ready.
- Scope: `HcclChannel._handle_init_msg` MemReg path only
  (`lmcache_ascend/v1/transfer_channel/hccl_channel.py`).

## Why

- On `HcclInitRequest`, the server starts `complete_handshake()` (background
  `accept()`), waits only for `accept_started_event`, then replies Init.
- Client may finish `connect()` and send MemReg before T3 writes the handle.
- `HcclReadyRequest` already waits on `_peer_ready_events`; MemReg did not.
- Fix: wait on the same `ready_event` (timeout 120s) before looking up
  `conn_handles_dict`, and raise clearly if the handle is still missing.

## Test plan

- [ ] Apply only this change.
- [ ] Launch Ascend PD disagg (HCCL, prefiller + decoder).
- [ ] Send a request through the disagg proxy.
- [ ] Inject delay right after accept().
- [ ] Decoder no longer logs MemReg `KeyError` / init-loop failure before
      “Connection established”.
- [ ] Prefiller no longer raises on `server_mem_handle_bytes`; KV transfer
      completes and proxy returns successfully.

## Notes

- Threads: T1 `lazy_init_peer_connection`, T2 `_init_loop` →
  `_handle_init_msg`, T3 nested `complete_handshake`.

### DIFF
--- a/lmcache_ascend/v1/transfer_channel/hccl_channel.py
+++ b/lmcache_ascend/v1/transfer_channel/hccl_channel.py
@@ -399,14 +399,18 @@ class HcclChannel(BaseMultiBufferChannel):
             # MemReg can arrive before that thread publishes conn_handle —
             # wait for the handshake ready event first.
             event = self._peer_ready_events.get(req.local_id)
-            if event is not None and not event.wait(timeout=120):
+            if event is None:
+                raise ConnectionError(
+                    f"No handshake initiated for peer {req.local_id}"
+                )
+            if not event.wait(timeout=120):
                 raise TimeoutError(
                     f"Timed out waiting for HCCL accept for peer {req.local_id}"
                 )
 
             with self._state_lock:
                 if req.local_id not in self.conn_handles_dict:
-                    raise KeyError(
+                    raise ConnectionError(
                         f"HCCL conn_handle missing for peer {req.local_id} "
                         "after handshake wait"
                     )
--- a/lmcache_ascend/v1/transfer_channel/hccl_channel.py
+++ b/lmcache_ascend/v1/transfer_channel/hccl_channel.py
@@ -400,10 +400,8 @@ class HcclChannel(BaseMultiBufferChannel):
             # wait for the handshake ready event first.
             event = self._peer_ready_events.get(req.local_id)
             if event is None:
-                raise ConnectionError(
-                    f"No handshake initiated for peer {req.local_id}"
-                )
-            if not event.wait(timeout=120):
+                raise ConnectionError(f"No handshake initiated for peer {req.local_id}")
+            if not event.wait(timeout=5):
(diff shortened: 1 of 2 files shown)


matthewygf (approved):
maybe 120s is a bit too long tho, wdyt @chloroethylene

lmcache_ascend/v1/transfer_channel/hccl_channel.py
```diff
@@ -394,9 +394,26 @@ def complete_handshake():
...
+            # Fix Github issue #263: HCCL handshake timeout due to race condition.
+            # accept() runs in a background thread started by HcclInitRequest.
+            # MemReg can arrive before that thread publishes conn_handle —
+            # wait for the handshake ready event first.
+            event = self._peer_ready_events.get(req.local_id)
+            if event is None:
+                raise ConnectionError(
+                    f"No handshake initiated for peer {req.local_id}"
+                )
+            if not event.wait(timeout=120):
```
matthewygf: Thinking about it, a timeout of 120s for a single request blocks other, since the socket is a REP. ITs worth lowering it down to 5s ? Since that is more than enough to wait on a slow peer I think ? @marcobarlo

matthewygf (changes requested):
Left a cmt.

lmcache_ascend/v1/transfer_channel/hccl_channel.py
marcobarlo: Agree. lgtm

tests/v1/transfer_channel/test_hccl_channel_write.py
```diff
@@ -765,6 +765,151 @@ def multi_buffer_receiver_process(
         sys.exit(1)
 
 
+# ──────────────────────────────────────────────────────────
+# Handshake race (issue #263): delay after accept() before publish
+# ──────────────────────────────────────────────────────────
```
matthewygf: maybe can remove these

matthewygf (changes requested):

matthewygf (approved):
lgtm, and tested

PR #270, opened 2026-07-28
[Bug] Restore normal-mode KV cache loading during retrieval

GrayMiao123:
## Summary

Restore KV cache loading in the normal retrieval path when
`save_only_first_rank` is disabled.

## Root cause

The Ascend `retrieve()` override handled KV loading inside the sharded
broadcast path, but omitted the normal-mode call to
`gpu_connector.batched_to_gpu()`.

As a result, LMCache reported the prefix as retrieved and returned a hit mask,
but the retrieved KV data was never loaded into vLLM's NPU KV cache. The model
then consumed stale or invalid KV data and generated corrupted output.

## Fix

Call `batched_to_gpu()` for retrieved chunks when
`save_only_first_rank` is disabled, while preserving the existing
sharded-broadcast behavior.

A regression test verifies that:

- retrieved memory objects are passed to `batched_to_gpu()`;
- chunk offsets and connector arguments are preserved;
- retrieved memory objects are released afterward.

### Unit test

```bash
python -m pytest tests/v1/test_cache_engine_retrieve.py -q
```
Result:

```text
1 passed
```

### DIFF
--- a/lmcache_ascend/v1/cache_engine.py
+++ b/lmcache_ascend/v1/cache_engine.py
@@ -681,6 +681,12 @@ class AscendLMCacheEngine(LMCacheEngine):
                 self._pipelined_sharded_broadcast_and_load(
                     reordered_chunks, ret_mask, **kwargs
                 )
+        elif len(reordered_chunks) > 0:
+            with retrieve_stats.profile_to_gpu():
+                _, memory_objs, starts, ends = zip(*reordered_chunks, strict=False)
+                self.gpu_connector.batched_to_gpu(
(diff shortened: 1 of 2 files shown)


matthewygf (approved):
Ah great ! Thanks for catching this !

GrayMiao123:
Cheers!

Just a note: the 910B/910C unit-test and performance jobs all failed at
`Checkout (Workflow Run)` before any tests were executed due to the fork
checkout protection in `actions/checkout`.

The code-quality check passed, and the local NPU validation passed as
described above.

matthewygf:
Flaky tests, need fixing, merging

PR #271, opened 2026-07-31
Dsv4 support

larksudo:


matthewygf:
I reviewed the current head (`df2dd9a`) and found two issues that look merge-blocking, plus one format-detection edge case:

1. **Connector constructs two LMCache managers and orphans the first manager’s services.** This is **not** caused by monkey-patching a class after an instance already exists. On the normal non-native path, the [import-time patch](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/__init__.py#L607-L623) first rebinds `LMCacheConnectorV1Impl` to `LMCacheAscendConnectorV1Impl`; only later does vLLM construct the outer connector.

   ```mermaid
   sequenceDiagram
       participant Patch as Import-time patch
       participant Factory as vLLM factory
       participant Child as Ascend connector
       participant Parent as Parent connector
       participant Runtime as LMCache runtime
       Patch->>Patch: Rebind implementation to Ascend impl
       Patch->>Factory: Register Ascend outer connector
       Factory->>Child: Construct connector
       Child->>Parent: super().__init__(...)
       Parent->>Runtime: Construct Impl A
       Runtime->>Runtime: Manager A + start services
       Parent-->>Child: _lmcache_engine = Impl A
       Child->>Runtime: Construct Impl B explicitly
       Runtime->>Runtime: Manager B + start services
       Runtime-->>Child: Impl B
       Child->>Child: _lmcache_engine = Impl B
       Note over Child,Runtime: No shutdown for Impl A / Manager A
   ```

   The duplicate construction is deterministic once this new outer connector is selected. The upstream parent [constructs and assigns Impl A](https://github.com/vllm-project/vllm/blob/v0.18.0rc1/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py#L84-L115). Every LMCache implementation construction [creates an `LMCacheManager` and calls `start_services()`](https://github.com/LMCache/LMCache/blob/v0.4.5/lmcache/integration/vllm/vllm_v1_adapter.py#L451-L477). The PR then constructs Impl B and replaces the same field in both the [in-process wrapper](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/integration/vllm/lmcache_ascend_connector.py#L26-L42) and [dynamic wrapper](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/integration/vllm/lmcache_ascend_connector_v1.py#L36-L54), without calling `Impl A.shutdown()` / `Manager A.stop_services()`.

   Strictly, this does not necessarily create two cache engines: `LMCacheEngineBuilder` normally returns the same engine singleton. The guaranteed problem is two manager/service initializations followed by an unclean overwrite. On workers, the first manager’s offload service [binds an IPC socket and starts a daemon thread in its constructor](https://github.com/LMCache/LMCache/blob/v0.4.5/lmcache/v1/offload_server/zmq_server.py#L22-L78), so the second construction can collide with the same endpoint or enter degraded mode while the first service remains outside the connector shutdown path.

   The explicit reconstruction is also unnecessary on the normal patched path: vLLM stores `kv_cache_config` on the parent, and the PR’s [multi-group constructor already falls back to `parent._kv_cache_config`](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/integration/vllm/multi_group_vllm_adapter.py#L490-L505). Please construct the implementation exactly once; if `use_native=True` must be supported, select the intended implementation once rather than create-and-replace it. Please also add a regression test asserting one manager initialization/start and one cleanup per public connector, and ensure the in-process wrapper delegates shutdown to its retained implementation.

2. **Rollback can create a discontinuous logical token history.** After a retrieval failure, vLLM resets `request.num_computed_tokens`; that is the authoritative logical-token rollback boundary. This path computes `new_token_ids` beginning at that boundary, but [then may truncate the existing tracker to a smaller block-derived capacity](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/integration/vllm/multi_group_vllm_adapter.py#L796-L824) before `new_block_ids` are merged. If the tracker is capped below `num_current_tokens`, `update()` appends tokens starting at `num_current_tokens`, leaving a gap in `token_ids` and corrupting subsequent chunk hashes/load-save keys.

   The current cap also mixes units: `num_blocks * block_size` counts physical compressed rows, whereas the tracker is counted in logical tokens. For a compressed group, logical capacity would be `num_blocks * block_size * compress_ratio` (for example, one 128-row block at ratio 8 covers 1,024 logical tokens). However, the safer fix is not to silently cap the logical history at all: restore exactly `request.all_token_ids[:num_current_tokens]`, clamp `num_saved_tokens` to the same boundary, and treat physical-capacity validation as an assertion/diagnostic after the block-table update. If a hard guard is retained, it must use per-group logical capacity and must not silently move the tracker below vLLM's rollback boundary. Please add a compressed-group retrieval-failure regression test that verifies the token history remains contiguous.

3. **Independent equal-block-size planes can be misclassified.** [`_is_multi_plane_tuple`](https://github.com/LMCache/LMCache-Ascend/blob/df2dd9ae76c2e4432117d8da0706e14db32227a8/lmcache_ascend/v1/kv_format.py#L35-L45) only recognizes independently paged planes when their block sizes differ. Independent planes with equal block sizes can therefore fall through to `SEPARATE_KV`/`MLA_KV`, and equal-shape pairs are later rejected by the multi-plane parameter builder. A generic equal-block-size multi-plane test would clarify and protect the intended behavior.

I recommend holding the merge until the first two issues are addressed. I also noticed that no CI/status checks are currently reported for this PR.

docker/mindspore/Dockerfile.310p.openEuler
matthewygf, line 1: Do we need the mindspore change here ? @larksudo

docker/mindspore/Dockerfile.a2.openEuler
matthewygf, line 1: do we need the mindspore change here ? @larksudo @chloroethylene

docs/deepseek_v4/dsv4-deployment-guide.md
```diff
@@ -0,0 +1,253 @@
...
+    --port 8900 \
+    --block-size 128 \
+    --compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'\
+    --async-scheduling \
+    --additional-config '
+    {"ascend_compilation_config":{
+        "enable_npugraph_ex":true,
+        "enable_static_kernel":false
+        },
+    "enable_flashcomm1": false,
```
matthewygf: question, we had flash_comm1 false here, but the envvar for VLLM_ASCEND_ENABLE_FLASHCOMM1=1 ?

docs/deployment.md
matthewygf, line 1: Not sure whether we need updates for Mindspore anymore. @larksudo @chloroethylene

matthewygf (changes requested):

matthewygf:
Follow-up on `cc713960`:

**New merge-blocker: this commit deletes modules that the current head still imports.** In particular:

- [`storage_backend/__init__.py`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/v1/storage_backend/__init__.py#L20) still imports `lmcache_ascend.v1.storage_backend.p2p_backend`, but `p2p_backend.py` is deleted.
- [`npu_connectors.py`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/v1/npu_connector/npu_connectors.py#L36-L41) still imports the deleted `proxy_memory_obj.py` and `transfer_context.py`.
- [`blend/utils.py`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/v1/blend/utils.py#L10) still imports the deleted `blend/blender.py`.

These are unconditional top-level imports, so the affected package paths now fail with `ModuleNotFoundError` before any DSv4 behavior can run. The commit removes about 3,000 lines across those four modules while its message only describes the manager/rollback/CI fixes. Please restore the files, or include and validate a complete migration of every importer to the intended replacement.

The rollback change itself is correct: restoring `request.all_token_ids[:num_current_tokens]` at vLLM's logical rollback boundary avoids the discontinuity, so this path no longer needs to derive a cap from block capacity or compression ratios. The dynamic wrapper also now constructs its implementation once.

One additional merge-blocking lifecycle path remains, though. The active in-process wrapper still double-initializes:

1. [`_patch_vllm_v1_adapter`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/__init__.py#L607-L625) rebinds the implementation to `LMCacheAscendConnectorV1Impl`.
2. [`_patch_vllm_ascend_connector`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/__init__.py#L643-L687) registers the in-process `LMCacheAscendConnector`.
3. That wrapper's [`super().__init__()`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/integration/vllm/lmcache_ascend_connector.py#L31-L38) therefore already constructs the patched Ascend impl/manager.
4. It then [constructs another impl and overwrites `_lmcache_engine`](https://github.com/LMCache/LMCache-Ascend/blob/cc713960d53c2ab673d91de31dcb2c82abbe26a6/lmcache_ascend/integration/vllm/lmcache_ascend_connector.py#L39-L44), again without shutting down the first manager's services.

Please make the in-process wrapper construct exactly once as well, and add a regression test asserting one manager start and one cleanup for each public wrapper. The equal-block-size independent-plane detection finding also remains unchanged.

The new CI triggers did start, but they did not execute tests. Both [unit-test jobs](https://github.com/LMCache/LMCache-Ascend/actions/runs/31085049275) and the [perf/accuracy job](https://github.com/LMCache/LMCache-Ascend/actions/runs/31085049686) fail in the initial custom “Set Status to Running” step: the fork-PR token has `statuses: read`, so the commit-status POST returns `403 Resource not accessible by integration`; every install/test step is then skipped. Please skip or make the custom status-writing steps non-fatal for `pull_request` events (the workflow job itself already reports a check), then rerun CI. Code Quality passed, but the runtime changes remain unvalidated.

lmcache_ascend/integration/vllm/vllm_v1_adapter.py
```diff
@@ -395,7 +709,30 @@ def request_finished(
...
+        # ``super().request_finished()`` had two bare asserts
+        # (``assert self.lmcache_engine is not None`` and
+        # ``assert self.lookup_client is not None``) that crash the
+        # EngineCore process on Scheduler-side aborts where the engine
+        # is not initialized.  Guard the super() call so that an old
+        # upstream version cannot kill the engine; if it raises, we
+        # perform the abort cleanup ourselves (fail-soft).
+        try:
+            _, return_params = super().request_finished(request, block_ids)
+        except AssertionError:
```
matthewygf: **Medium — this can cancel the same backend request twice after a partially executed `super()` call.**

In LMCache v0.4.5, the aborted-request path calls `storage_manager.cancel_request(request_id)` *before* checking `assert self.lookup_client is not None`. Therefore, when the engine exists, `async_loading=True`, and the lookup client is missing, `super().request_finished()` has already performed the storage cancellation when it raises. This handler then calls `_local_abort_cleanup()`, which performs the same cancellation again. For `PDBackendAsync`, each call schedules an `_abort_request()` coroutine, so the two tasks can race and send duplicate `CancelNotif`s before the first task clears its request state.

Please decide whether to use the compatibility path from the dependency state *before* calling `super()` (engine missing, or async lookup client missing), and run each available cleanup exactly once. The fallback should also preserve upstream's output contract rather than returning a copy of all input `kv_transfer_params`. Regression cases should cover both combinations: (1) engine missing/client present, and (2) engine present/client missing, asserting one cancellation per available component.

matthewygf (commented):
The compatibility guard fixes the common Scheduler-side assertion, but the fallback needs to be chosen before invoking a side-effectful upstream method.

larksudo:
Concern about double-initializes: this connector is really necessary for non-DSv4 paths — it's the entry point there. Removing it breaks import lmcache_ascend and general deployments.

matthewygf:
Re-reviewed the updated head `289d475`. The two new commits resolve the earlier merge blockers I could verify:

- the four unconditionally imported modules are restored;
- the in-process connector now creates only the implementation constructed by `super().__init__()`, and `shutdown()` delegates to that implementation, so the duplicate manager/service lifecycle is removed;
- rollback keeps the authoritative logical-token boundary;
- `request_finished()` now performs at most one storage cancellation and one lookup cancellation, with targeted scheduler-role tests.

One correctness edge remains in the generic multi-plane path. `build_flat_kv_caches()` bundles every non-shared multi-plane entry when its tensors belong to separate scheduler groups, but `_is_multi_plane_tuple()` recognizes that bundle only when the tensors have different block sizes. If two independent planes both use (for example) block size 128, the bundle is still created, but `KVCacheFormat.detect()` can classify it as `SEPARATE_KV` (or, for some four-plane shapes, `DSA_C8_KV`) instead of `MULTI_PLANE_KV`. The connector then does not build the per-plane pointer/block-size parameters, even though the planes have independent scheduler mappings.

Could we either carry explicit “independent bundled planes” provenance into format detection (preferred), or otherwise distinguish equal-block independent planes from DSA-C8/shared-storage tuples? A regression test with two independently allocated, equal-block-size planes assigned to different scheduler groups would pin this down.

CI note: Code Quality passes. The current Unit Tests failure does not appear attributable to these commits: the 910B job reported three failures in unchanged P2P tests while `p2p_backend.py` is identical to the PR base, then the job was cancelled before pytest emitted failure traces; the 910C job failed downloading `actions/checkout@v4` with HTTP 429. Please rerun those jobs. `git diff --check` also still reports trailing whitespace in `examples/offload.py` and `tests/v1/test_npu_connector.py`.

tests/v1/storage_backend/test_ascend_p2p_backend.py
```diff
@@ -1659,14 +1781,13 @@ async def slow():
...
         backend._cleanup_late_sync_get_result = (
-            lambda done, lookup_id, operation, unpin, timeout_at=None: (
+            lambda done, lookup_id, operation, unpin: (
                 AscendP2PBackend._cleanup_late_sync_get_result(
                     backend,
                     done,
                     lookup_id,
                     operation,
                     unpin,
-                    timeout_at=timeout_at,
```
larksudo, line 1669: _cleanup_late_sync_get_result does not have this parameter

matthewygf:
Post-merge follow-up on `5ac6b78`: the equal-block provenance fix is sound in the config-verified branch, but the fallback currently applies `MultiPlaneBundle` more broadly than its documented invariant.

A standard `SEPARATE_KV` entry can reproduce this:

1. The model has multiple scheduler groups overall, so bundling is enabled.
2. One layer belongs to a single scheduler group `g` and exposes a normal independently allocated `(K, V)` pair with the same block size/shape.
3. `ordered_scheduler_groups_for_layer()` maps the two tensors to `[g, g]`.
4. `_planes_match_config_groups()` correctly returns false because there is only one containing group.
5. `_is_kernel_native_tuple()` also returns false because `SEPARATE_KV` is not in `_KERNEL_NATIVE_FORMATS`.
6. The fallback at `multi_spec_flatten.py:307-313` nevertheless wraps the pair in `MultiPlaneBundle`.
7. `KVCacheFormat.detect()` trusts that tag unconditionally and returns `MULTI_PLANE_KV`, routing an ordinary shared-index K/V pair through packed multi-plane storage/kernel handling.

This contradicts the tag's own contract that it marks bundles “after config verification.” The narrow fix is to keep the fallback as a plain `tuple(planes)`; heterogeneous fallback bundles are already recognized by the existing block-size heuristic. Only the config-verified branch at lines 289-301 should attach `MultiPlaneBundle`.

Please add a regression with a multi-group config where the tested layer appears in only one group and its entry is an equal-shape `(K, V)` pair. The expected result is an untagged tuple and `KVCacheFormat.SEPARATE_KV`, while the new two-distinct-group equal-block tests should remain `MULTI_PLANE_KV`.

PR #273, opened 2026-08-07
[PD][P2P] Allow PD and P2P to be used together on Ascend

shengyan-chen:
## Overview

This PR enables Ascend PD and P2P to be used together on the PD sender path and hardens the related request lifecycle handling for PD KV transfer.

The changes cover three correctness areas:
- config validation for `enable_pd=true` + `enable_p2p=true` on Ascend sender
- PD/P2P store correctness when some destination keys already exist
- request-scoped PD receiver leases, delay-pull proxy isolation, and vLLM preemption cleanup

## Impact

Before this PR:
- Ascend config validation rejected `enable_pd=true` and `enable_p2p=true` even for the sender role.
- `allocate_and_copy_objects` could return misaligned keys when earlier keys were skipped because they already existed in the target backend, causing newly copied objects to be paired with the wrong keys.
- A shared PD receiver hit could be consumed or released through one request while another request still depended on the same PD entry.
- vLLM 0.23 preemption metadata was not carried through the Ascend connector path, so preempted requests could miss LMCache cleanup.

These issues affect PD/P2P mixed deployment correctness, especially when sender-side local/P2P cache hits, PD handoff, receiver-side shared hits, delay-pull proxy objects, and vLLM preemption interact.

## What changed

Config and sender-side store path:
- Allow `enable_pd=true` and `enable_p2p=true` for `pd_role="sender"` with explicit Ascend validation.
- Keep upstream PD validation while preserving Ascend-specific P2P requirements.
- Clamp PD producer store skip to `disagg_spec.num_transferred_tokens`, so the producer only skips tokens already transferred to the decoder.
- Preserve key/object alignment in `allocate_and_copy_objects` by tracking `allocated_keys` together with allocated memory objects.

PD receiver lifecycle:
- Add receiver-side `PDEntry` ownership metadata.
- Track PD hits by request id through lookup/retrieve context.
- Add request-scoped PD leases and release them from `lookup_unpin` / `retrieve`.
- Create request-local `ProxyMemoryObj` clones for delay-pull hits so one request consuming a proxy does not poison shared hits for another request.
- Delay PD Done signaling until the last active request lease is released.

vLLM preemption handling:
- Carry `scheduler_output.preempted_req_ids` in Ascend connector metadata for vLLM 0.23.
- Keep compatibility with the legacy vLLM 0.18 set-based preemption payload.
- On preemption, release lookup pins for preempted request ids.
- For async producer workers, drain pending stores before KV blocks can be reused.

## Tests

Added and adjusted regression tests for:
- key/object alignment after skipped existing keys
- PD producer skip clamping to decoder-transferred tokens
- PD request lease routing from `StorageManager.batched_contains`
- shared PD receiver entries surviving until the last request lease is released
- request-local delay-pull proxy leases
- PD transfer context Done signaling after the last request lease
- vLLM 0.23 metadata payload and legacy set payload preemption handling
- async producer preemption cleanup and drain behavior

## Scope note

This PR focuses on the first PD/P2P lifecycle boundary needed for Ascend sender compatibility and receiver request cleanup.

It does not include later PD KVPool work such as DSA KV layout changes, token-aware disaggregated proxy scheduling, or additional delay-pull handoff race diagnostics. Those will be handled in follow-up PRs.

### DIFF
--- a/lmcache_ascend/__init__.py
+++ b/lmcache_ascend/__init__.py
@@ -228,6 +228,7 @@ def _patch_config():
         "When exceeded, the oldest entry is evicted to prevent unbounded "
         "memory growth. Default is 0 (unlimited).",
     }
+
     def _ascend_validate_config(self):
         if not (self.enable_pd and self.enable_p2p):
             return lmcache.v1.config._validate_config(self)
@@ -235,7 +236,7 @@ def _patch_config():
         if self.pd_role != "sender":
             raise ValueError(
                 "Invalid LMCache-Ascend config: `enable_pd=true` and "
-                "`enable_p2p=true` are only supported for `pd_role=\"sender\"` "
+                '`enable_p2p=true` are only supported for `pd_role="sender"` '
                 f"in Phase 1, got pd_role={self.pd_role!r}."
             )
 
@@ -390,10 +391,10 @@ def _patch_storage_manager():
         allocate_and_copy_objects as ascend_allocate_and_copy_objects,
     )
     from lmcache_ascend.v1.storage_backend.storage_manager import (
-        batched_get as ascend_batched_get,
+        batched_contains as ascend_batched_contains,
     )
     from lmcache_ascend.v1.storage_backend.storage_manager import (
-        batched_contains as ascend_batched_contains,
+        batched_get as ascend_batched_get,
     )
     from lmcache_ascend.v1.storage_backend.storage_manager import get as ascend_get
     from lmcache_ascend.v1.storage_backend.storage_manager import (
--- a/lmcache_ascend/__init__.py
+++ b/lmcache_ascend/__init__.py
@@ -228,9 +228,83 @@ def _patch_config():
         "When exceeded, the oldest entry is evicted to prevent unbounded "
         "memory growth. Default is 0 (unlimited).",
     }
+    def _ascend_validate_config(self):
+        if not (self.enable_pd and self.enable_p2p):
+            return lmcache.v1.config._validate_config(self)
+
+        if self.pd_role != "sender":
+            raise ValueError(
+                "Invalid LMCache-Ascend config: `enable_pd=true` and "
+                "`enable_p2p=true` are only supported for `pd_role=\"sender\"` "
+                f"in Phase 1, got pd_role={self.pd_role!r}."
+            )
+
+        if self.enable_async_loading:
+            raise ValueError(
+                "Invalid LMCache-Ascend config: `enable_async_loading=true` is "
+                "not supported when `enable_pd=true` and `enable_p2p=true` in "
+                "Phase 1. Use blocking retrieve with enable_async_loading=false."
+            )
+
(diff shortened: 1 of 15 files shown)


matthewygf:
I found two merge-blocking issues in the new PD/P2P lifecycle path:

### 1. Partial push allocation rollback leaves released objects addressable through `_pd_entries`

The new [`put()`](https://github.com/LMCache/LMCache-Ascend/blob/870b4c10ec09050417856132c42253473d8b44e4/lmcache_ascend/v1/storage_backend/pd/backend.py#L326-L336) records every object in both `data` and `_pd_entries`, and the new lookup/lease paths use `_pd_entries` as the authoritative map.

However, if a later allocation in the same push request fails, [the rollback loop](https://github.com/LMCache/LMCache-Ascend/blob/870b4c10ec09050417856132c42253473d8b44e4/lmcache_ascend/v1/storage_backend/pd/receiver_mixin.py#L106-L113) only executes `self.data.pop(k, None)`, then `release_memory_objects(allocated_objs)`. The matching `PDEntry` survives, pointing at an object whose reference was already released. A later `contains()`, `get_blocking()`, or request lease can therefore return freed/recycled storage, and later cleanup may release it again.

The tests currently hide this: `_make_pd_backend_stub` replaces production's plain `data` dict with a custom `_PDDataDict.pop()` that also removes `_pd_entries`. Production does not have that coupling.

Please centralize rollback through `_delete_pd_entry_locked` (without double-releasing before the existing `release_memory_objects` call), or add a helper that atomically removes both indexes. A regression test should keep `data` and `_pd_entries` as independent dictionaries, force failure after at least one successful allocation, and assert that neither map retains the released object.

### 2. The newly accepted combined configuration can fail immediately during backend creation

[`_ascend_validate_config`](https://github.com/LMCache/LMCache-Ascend/blob/870b4c10ec09050417856132c42253473d8b44e4/lmcache_ascend/__init__.py#L232-L303) accepts an `enable_pd=true`, `enable_p2p=true`, `pd_role="sender"` configuration without requiring `local_cpu=true` or a positive local CPU capacity. This matters because the existing PD sender example uses `local_cpu: false`.

During construction, [`CreateStorageBackends`](https://github.com/LMCache/LMCache-Ascend/blob/870b4c10ec09050417856132c42253473d8b44e4/lmcache_ascend/v1/storage_backend/__init__.py#L84-L121) skips creating `LocalCPUBackend` when PD is enabled and `local_cpu` is false, then the P2P branch executes `assert local_cpu_backend is not None`. Thus a configuration that the new validator explicitly accepts can abort at startup before either backend is usable.

Please either validate that mixed sender mode requires `local_cpu=true` and `max_local_cpu_size > 0`, or make backend creation provision the LocalCPU backend whenever P2P is enabled. A test should validate and construct backends from a production-like mixed sender config; the current patch has no combined-config construction test.

larksudo:
Do not support InProcess in main branch anymore.

PR #274, opened 2026-08-07
Fix ProxyMemoryObj lookup pin lifecycle

### LINKED ISSUE #250: [BUG] PD disaggregation shared-prefix KV reuse can cause 0-hit and stale proxy lifecycle issues
**Describe the bug**  
In PD disaggregated mode with shared prefixes, the decoder may report `0` LMCache hit tokens or inconsistent lookup results across TP ranks. The issue is caused by key-level `already_sent` reuse without request-level ownership tracking. A later request can skip KV transfer because the key already exists on the decoder, but that key may be removed after an earlier request retrieves it.

In delay-pull mode, the problem is more severe because the decoder stores `ProxyMemoryObj` remote references instead of real local KV. Reusing an existing proxy across requests can bind a request to another request’s remote pinned buffer lifecycle.

**To Reproduce**  
1. Run LMCache-Ascend PD disaggregated prefill/decode with shared prefix enabled, e.g. `shared-system-prompt=5k`.
2. Use sync push or delay-pull mode with high QPS / concurrent requests.
3. Observe decoder LMCache logs.
4. See logs such as:
   - `LMCache hit tokens: 0, need to load: 0`
   - `Lookup results (number of hit tokens) differ across (TP and PP) ranks`

**Expected behavior**  
Requests with shared prefixes should consistently hit LMCache across all TP ranks. Reusing an existing key should be safe only when the decoder correctly tracks all request owners and releases the underlying KV/proxy only after the last owner has finished.

**Screenshots**
test config：

Image

logs：
Image

Image

Futuresxy:
## Summary

This is a partial fix for #250. It addresses only the temporary lookup-pin imbalance in `_partition_keys()`: releasing an already-present `ProxyMemoryObj` returned by a pinned lookup must not consume the proxy's base transfer ownership.

`ProxyMemoryObj.ref_count_up()` now records a balanced logical reference, and `ref_count_down()` only decrements the shared transfer context when the base owner is actually released. This preserves the existing discard and consumed-proxy semantics while preventing `_partition_keys()` / `release_memory_objects()` from sending Done early for this lookup path.

## Scope

This PR intentionally does not close #250.

Concurrent retrieve requests can still receive the same unconsumed `ProxyMemoryObj` from `PDBackend.data`. `batched_to_gpu()` mutates the shared `_backing_obj`, `set_backing_obj()` resets `_resolved`, and there is currently no consumer-claim or single-flight mechanism. The first request to call `mark_consumed()` / `send_done_now()` can therefore release the remote lease before another request using that proxy has finished.

That request-level ownership redesign remains tracked in #250 and is outside this focused PR.

Related #250.

## Tests

- `PYTHONPATH=$PWD/test_bootstrap:$PWD:$PWD/../LMCache-v0.4.4 /data/pyptouser/songxiaoyou/va-venv/bin/python -m pytest --noconftest tests/v1/storage_backend/test_proxy_memory_obj.py -q`
- `LMCACHEPATH=$PWD/../LMCache-v0.4.4 PYTHONPATH=$PWD/test_bootstrap:$PWD:$PWD/../LMCache-v0.4.4 /data/pyptouser/songxiaoyou/va-venv/bin/python -m pytest --noconftest tests/v1/storage_backend/test_ascend_pd_backend.py -k 'partition_keys_proxy or partition_keys or pull_delay_transfer_context_done_callback_is_idempotent' -q`
- `LMCACHEPATH=$PWD/../LMCache-v0.4.4 PYTHONPATH=$PWD/test_bootstrap:$PWD:$PWD/../LMCache-v0.4.4 /data/pyptouser/songxiaoyou/va-venv/bin/python -m pytest --noconftest tests/v1/storage_backend/test_ascend_p2p_backend.py -k 'proxy_ref_count_down or proxy_mark_consumed or proxy_shared_context' -q`

Also smoke-tested `torch_npu` through the local task queue on Ascend 910/910C-class hardware.

### DIFF
--- a/lmcache_ascend/v1/proxy_memory_obj.py
+++ b/lmcache_ascend/v1/proxy_memory_obj.py
@@ -15,6 +15,7 @@ before scattering, enabling overlap of remote fetch and NPU scatter operations.
 # Standard
 from typing import Any, List, Optional
 import asyncio
+import threading
 
 # Third Party
 from lmcache.logging import init_logger
@@ -88,6 +89,11 @@ class ProxyMemoryObj(MemoryObj):
         self._resolved = False
         self._consumed = False  # True after data scattered into KV cache
         self._released = False
+        self._ref_lock = threading.Lock()
+        # One base owner represents the proxy's transfer-context ownership.
+        # Temporary lookup pins increment this count but must not consume the
+        # base owner when they are released.
+        self._ref_count = 1
 
         # Store allocation metadata for deferred buffer operations
         if backing_obj is not None:
@@ -133,8 +139,10 @@ class ProxyMemoryObj(MemoryObj):
         sender's pinned memory may be released, so the proxy's remote
         buffer references are no longer valid.
         """
-        self._consumed = True
-        self._released = True
+        with self._ref_lock:
+            self._consumed = True
+            self._released = True
+            self._ref_count = 0
 
     @property
     def backing_obj(self) -> Optional[MemoryObj]:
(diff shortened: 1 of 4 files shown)


matthewygf:
Thanks — this looks correct as a narrow fix for the `_partition_keys()` temporary-pin imbalance, and I do not think it should be blocked on the broader request-ownership redesign.

Could you please make one small scope/documentation updates before merging?

1. Update the `_lookup()` [docstring](https://github.com/LMCache/LMCache-Ascend/blob/148a47aca86af28cc47f18ca31decd68e37436ff/lmcache_ascend/v1/storage_backend/pd/backend.py#L259-L271), which still says `ProxyMemoryObj.ref_count_up/down()` are no-ops.

Futuresxy:
Addressed, thanks. I updated the `_lookup()` docstring in `9b66f49` to describe the proxy's temporary logical lookup references above its base transfer owner. I also updated the PR description to state explicitly that this is only a partial fix for #250 and that #250 remains open. The concurrent-consumer claim/single-flight and request-level remote-lease ownership problem is intentionally left out of this focused PR and remains tracked in #250.

matthewygf (approved):
lgtm! Thanks for the contribution!

Issue #277, opened 2026-08-21
[BUG] Recomputation causes severe precision deterioration.


OceanAnn:
When I set the recomputationrate between 0 and 1, the accuracy on my dataset drops significantly, and the higher the recalculation rate, the lower the accuracy. However, when set to 1 or 0, it is normal.

**Platform & Software (please complete the following information):**
 - OS: [openEuler]
 - Hardware: [910B]
 - vLLM Version: [v0.23.0]

zihanlin-ai:
@OceanAnn Two questions, please:

1. Was `enable_prefix_caching` left on? It is vLLM's default, and `examples/blending/README.md` lists it as incompatible with CacheBlend — but nothing enforces that.
2. Could you post your full blend config and the model?

Why: when vLLM's prefix cache already covers part of a request, `start_load_kv` masks off the leading tokens, and `LMCBlender.process_qkv` then computes the deviation over `k[num_falses:]` (blend-buffer coordinates) but applies the resulting `top_indices` to the full-length `k`, `v`, `q`, `residual` and `positions` (sequence coordinates). Every recomputed token picks up the QKV and the RoPE position of the token `num_falses` slots earlier. The damage scales with the number of recomputed tokens, and ratio 0 never enters the path — which matches your report, except that it should hurt at ratio 1 too, so I would like to confirm your config before calling this the cause.

Relative KV error against a full prefill, `num_falses=64` of 256 tokens (float32 reimplementation of the path):

| ratio | 0.05 | 0.15 | 0.50 | 1.00 |
| --- | --- | --- | --- | --- |
| as shipped | 0.47 | 0.55 | 0.82 | 1.13 |
| indices offset by `num_falses` | 0.40 | 0.35 | 0.23 | 0.27 |

Draft PR #282 makes that configuration fail loudly instead of silently producing shifted KV.

Ruled out on 910B2, each against a float32 reference: the FIA mask with `Q < K` (0.004 rel), `npu_mrope` with scattered positions (0.008), `rotary_embedding_k_fused` with per-chunk position shifts (0.0156), and the blend algorithm itself — on real Qwen3-4B weights it converges monotonically to the full-prefill KV as the ratio rises.

PR #282, opened 2026-08-29
[Bug] Reject Cacheblend + vLLM prefix caching at startup

zihanlin-ai:
## Summary

The legacy in-process CacheBlend path used by LMCache-Ascend is not compatible
with vLLM prefix caching. After a prefix-cache hit, LMCache retrieves only the
remaining suffix into the blend buffer, while `LMCBlender.process_qkv` applies
suffix-local indices to full-sequence tensors. This can corrupt QKV selection
and positions.

This PR rejects the unsupported combination in `_apply_extra_config`, after
LMCache extra config is applied but before `LMCacheManager` and its services are
created. Other configurations are unchanged. Users of this connector must set
`enable_prefix_caching=False` when enabling CacheBlend.

## Validation

- 910B2 / CANN 9.1.0 / torch_npu 2.10.0: `tests/integration` — 19 passed.
- `test_request_finished_chunk_hashes` fails identically on `main` and this PR.
- All 5 focused guard cases pass; both rejection cases fail on `main` as a
  negative control.
- Lint, format, type, and syntax checks pass.

Upstream context: LMCache/LMCache#3229 reproduces the combined legacy path
failing on GPU, and the upstream legacy example disables prefix caching.

Related to #277, but not claimed as its fix because #277 was not reproduced end
to end and its reported ratio-1 behavior does not match this failure mode.

### DIFF
--- a/lmcache_ascend/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache_ascend/integration/vllm/vllm_v1_adapter.py
@@ -43,6 +43,28 @@ class LMCacheAscendConnectorV1Impl(LMCacheConnectorV1Impl):
         self._late_finished_sending: set[str] = set()
         logger.debug("store_async: %s", self.store_async)
 
+    def _apply_extra_config(self, config, vllm_config: "VllmConfig") -> None:
+        """Apply vLLM extra config, then reject unsupported combinations.
+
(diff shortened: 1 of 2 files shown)


matthewygf (approved):
Thanks for the contribution !

PR #281, opened 2026-08-29
[Feature] Map torch.int8 to AscendType::INT8

zihanlin-ai:
Fixes #31

## What changed

Map `at::ScalarType::Char` (`torch.int8`) to `AscendType::INT8` in
`get_dtype_from_torch`. `torch.uint8` remains unsupported, and the RoPE path
keeps its existing float-only validation.

Because the helper is not exposed through pybind, tests cover the public
`multi_layer_kv_transfer` and `single_layer_kv_transfer` operations with int8
round trips. The paged-cache test utility now generates integer data as well.

## Validation

Environment: 910B2, CANN 8.5.1, torch_npu 2.9.0.post1.

| Test | Current main | This PR |
|---|---:|---:|
| `test_mem_kernels.py -k int8` | 12 failed: unsupported dtype | 12 passed |
| `test_mem_kernels.py` + `test_pos_kernels.py` | 341 passed | 353 passed |

An earlier contributor expressed interest in #31 in November 2025 but did not
open a PR; happy to coordinate if needed.

PR #284, opened 2026-09-01
[csrc] Extend multi_layer_kv_transfer bindings; fix cann 9.1.0 build; map int8/uint8

chloroethylene:
## Why

Kernel-side prerequisite for the MP-mode fused-transfer port (first of a
4-PR series; the Python-side changes follow separately). Landing the csrc
layer alone keeps it independently verifiable: every change is backward
compatible for existing callers.

## What

- `multi_layer_kv_transfer` / `fused_multi_layer_kv_transfer` gain trailing
  `dsa_c8_scale_plane_bytes` / `paged_kv_block_size` parameters (py::arg
  names, zero defaults) so keyword-arg kernel calls bind; existing
  positional callers are unaffected.
- Drop the standalone `multi_plane` entry point and the DSA_C8 branch
  (DSA_C8 now fails fast with a `TORCH_CHECK`); no Python caller references
  either.
- `get_dtype_from_torch` maps both `torch.int8` (Char, #281) and
  `torch.uint8` (Byte, needed for DSA scale planes) to
  `AscendType::INT8` — both are 1-byte types the kernels treat as raw
  bytes. (Rebase note: #281 landed upstream while this was in flight; this
  keeps both mappings.)
- cann 9.1.0 build fixes: hcomm v9.1.0 include moves, hixl pkg_inc, hcomm
  submodule bump; kvcache-ops submodule bump for the multi-plane kernel.

## Verification

910B, against the exact main + lmcache v0.4.4 CI combination: full
`tests/v1/test_mem_kernels.py` — **344/344 parametrized cases pass**,
including the int8 tests added by #281.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/csrc/mem_kernels.cpp
+++ b/csrc/mem_kernels.cpp
@@ -3,6 +3,7 @@
 #include "utils.h"
 #include <ATen/ATen.h>
 #include <Python.h>
+#include <acl/acl.h>
 #include <pybind11/pybind11.h>
 #include <torch_npu/csrc/core/npu/NPUStream.h>
 #include <torch_npu/csrc/framework/OpCommand.h>
@@ -38,13 +39,15 @@ void multi_layer_kv_transfer(
     const torch::Device &paged_memory_device, const int page_buffer_size,
     const bool direction, const bool use_mla, const int kvcache_format_raw,
     const int64_t k_hidden_dims, const int64_t v_hidden_dims,
-    const int64_t dsa_hidden_dims) {
+    const int64_t dsa_hidden_dims, const int64_t dsa_c8_scale_plane_bytes,
+    const int32_t paged_kv_block_size) {
   uint8_t *key_value_ptr = get_kernel_ptr<uint8_t, torch::Tensor>(key_value);
 
   MultiLayerKVConfig config = prepare_multi_layer_kv_config(
       key_value, key_value_ptrs, slot_mapping, paged_memory_device,
       page_buffer_size, direction, use_mla, kvcache_format_raw, k_hidden_dims,
-      v_hidden_dims, dsa_hidden_dims);
+      v_hidden_dims, dsa_hidden_dims, dsa_c8_scale_plane_bytes,
+      paged_kv_block_size);
 
   // Calculate UB buffer parameters
   compute_multi_layer_ub_params(config, key_value, paged_memory_device,
@@ -78,7 +81,8 @@ void fused_multi_layer_kv_transfer(
     const bool direction, // true: from_gpu, false: to_gpu
     const bool use_mla, const int kvcache_format_raw,
     const int64_t k_hidden_dims, const int64_t v_hidden_dims,
-    const int64_t dsa_hidden_dims) {
+    const int64_t dsa_hidden_dims, const int64_t dsa_c8_scale_plane_bytes,
+    const int32_t paged_kv_block_size) {
   // get host cpu buffer pointer for aclrtMemcpyAsync
   uint8_t *key_value_ptr = static_cast<uint8_t *>(key_value.data_ptr());
   uint8_t *staging_cache_ptr =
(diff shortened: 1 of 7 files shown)


matthewygf:
**Merge blocker — DSA_C8 still reaches the non-fused kernel.**

`prepare_multi_layer_kv_config()` accepts raw format `5` and returns a config with `is_dsa_c8 = true`, but `multi_layer_kv_transfer()` then unconditionally queues `multi_layer_kv_transfer_kernel_v2`. The fail-fast `TORCH_CHECK(!config.is_dsa_c8, ...)` exists only in `fused_multi_layer_kv_transfer()`.

This leaves the public non-fused binding on the exact standard-kernel path that the fused-path comment warns can corrupt data:

- [non-fused dispatch](https://github.com/LMCache/LMCache-Ascend/blob/bd6ee7d0ff0ffc722ebfd5a74f3b216ea755616d/csrc/mem_kernels.cpp#L35-L71)
- [fused-only guard](https://github.com/LMCache/LMCache-Ascend/blob/bd6ee7d0ff0ffc722ebfd5a74f3b216ea755616d/csrc/mem_kernels.cpp#L134-L138)
- [format-5 configuration](https://github.com/LMCache/LMCache-Ascend/blob/bd6ee7d0ff0ffc722ebfd5a74f3b216ea755616d/csrc/utils.cpp#L33-L77)

Please reject DSA_C8 in the non-fused entry point before any handler is queued (or explicitly route it to a supported implementation). Add a regression that calls **both** Python bindings with format `5` and verifies each throws before launch. The reported 344 existing kernel cases do not cover this format.

csrc/mem_kernels.cpp
```diff
@@ -99,33 +104,43 @@ void fused_multi_layer_kv_transfer(
...
   }
 
   TORCH_CHECK(staging_cache_size >= required_size,
               "staging_cache size insufficient: need ", required_size,
               " bytes, got ", staging_cache_size);
 
+  // DSA_C8 needs the separate multi-plane kernel, which is not part of this
+  // MP-mode-only build; reject it explicitly instead of corrupting data via
+  // the standard kernel.
+  TORCH_CHECK(!config.is_dsa_c8,
```
marcobarlo, line 137: Why the PR extends the signature to accept C8 scale plane bytes and then forces a fail here?

csrc/mem_kernels.cpp
```diff
@@ -143,13 +158,15 @@ void fused_multi_layer_kv_transfer(
...
 
     // Step 2: Kernel (Gather or Scatter)
-    kvcache_ops::multi_layer_kv_transfer_kernel_v2(
-        dtype_num, slot_num, config.kvcache_format, config.aiv_num,
-        config.stream, config.page_buffer_ptrs, staging_cache_ptr,
-        config.slot_mapping_ptr, config.hidden_dims, config.kv_size,
-        config.num_layers, config.page_buffer_size, config.num_tokens,
-        config.singlePerLoopBuffer, config.maxTokensPerLoop, config.direction,
-        config.k_hidden_dims, config.v_hidden_dims, config.dsa_hidden_dims);
+    {
```
marcobarlo, line 161: Why do we need scoping?

csrc/mem_kernels.h
```diff
@@ -71,7 +71,9 @@ void multi_layer_kv_transfer(
     const torch::Device &paged_memory_device, const int page_buffer_size,
     const bool direction, const bool use_mla, const int kvcache_format_raw,
     const int64_t k_hidden_dims = 0, const int64_t v_hidden_dims = 0,
-    const int64_t dsa_hidden_dims = 0);
+    const int64_t dsa_hidden_dims = 0,
+    const int64_t dsa_c8_scale_plane_bytes = 0,
```
marcobarlo, line 75: DSA_C8 cannot be copied with the older multi layer copy kernel v2, the kernel only accepts one dtype for all the tensors to be copied

marcobarlo:
I propose to split this PR in two different PRs: 
1) to fix cann 9.1.0 build 
2) to extend the signatures (if needed) for tested model architectures (i.e., models with MHA/GQA, models with MLA, models with DSA, models with DSA C8, and models with hybrid KV ). 
In general, I didn't fully get why do we need to extend the signature with dsa_c8_scale_plane_bytes if the copy kernel cannot support it.

marcobarlo (changes requested):
I propose to split this PR in two different PRs:

1)    to fix cann 9.1.0 build

2)   to extend the signatures (if needed) for tested model architectures (i.e., models with MHA/GQA, models with MLA, models with DSA, models with DSA C8, and models with hybrid KV ).

In general, I didn't fully get why do we need to extend the signature with dsa_c8_scale_plane_bytes if the copy kernel cannot support it.

PR #289, opened 2026-09-03
fix(c_ops): add missing <stdexcept> include in mem_alloc.cpp

pjh456:
## Summary

Fix compilation of `c_ops` on newer toolchains.

`csrc/common/mem_alloc.cpp` uses `std::runtime_error` but never includes
`<stdexcept>`; it previously compiled only because another header pulled the
include in transitively. With GCC 13.3 / newer libstdc++ this no longer
happens and the build fails with:

```
error: 'runtime_error' is not a member of 'std'
note: 'std::runtime_error' is defined in header '<stdexcept>'; did you
      forget to '#include <stdexcept>'?
```

## Changes

- `csrc/common/mem_alloc.cpp`: add `#include <stdexcept>`.

## Test

- Build verified on aarch64 with CANN 9.1.0 (SOC Ascend910B4); `c_ops`
  compiles and links successfully.

### DIFF
--- a/csrc/common/exception.h
+++ b/csrc/common/exception.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <sstream>
+#include <stdexcept>
 #include <string>
 
 #if defined(USE_MINDSPORE) && (USE_MINDSPORE)
--- a/csrc/common/mem_alloc.cpp
+++ b/csrc/common/mem_alloc.cpp
@@ -5,6 +5,7 @@
 #include <cstring> // for strerror
 #include <errno.h>
 #include <numaif.h>
+#include <stdexcept>
 #include <string>
 #include <sys/mman.h>
 
--- a/csrc/mindspore/framework_hal.cpp
+++ b/csrc/mindspore/framework_hal.cpp
@@ -2,6 +2,7 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 #include <sstream>
+#include <stdexcept>
 #include <string>
 
 namespace py = pybind11;
--- a/csrc/mindspore/mem_kernels.cpp
+++ b/csrc/mindspore/mem_kernels.cpp
@@ -9,6 +9,7 @@
 #include <Python.h>
 #include <pybind11/numpy.h>
 #include <pybind11/pybind11.h>
+#include <stdexcept>
 #include <string>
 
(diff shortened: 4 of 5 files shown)


chloroethylene:
LGTM — confirmed the bug is real. 👍

Verified the full causal chain on a local checkout:

- mem_alloc.cpp throws std::runtime_error at 9 call sites with no <stdexcept> in its include set.
- It currently compiles only via managed_mem.h → <shared_mutex> → <bits/std_mutex.h> → <system_error> → <stdexcept>. GCC 13+ removed <system_error> from bits/std_mutex.h (throws now go through the declaration-only __throw_system_error in bits/functexcept.h), and no other header in this TU's include set reaches <stdexcept> on GCC 13. So GCC 13.x breaks exactly as described — the fix is correct and minimal.
 
One suggestion while you're at it: the same latent pattern exists in another file that also use std::runtime_error without including <stdexcept>:
  
- csrc/pybind.cpp (same c_ops target — likely just masked by torch/pybind11 headers today) 

They may happen to compile via transitive includes for now, but that's the same fragile dependency this PR fixes. Happy to see them land here or as a follow-up — either way this one is ready to merge.

pjh456:
@chloroethylene  Thanks for the thorough review! I've addressed the suggestion in commit 73c3da7:

- `csrc/pybind.cpp`: added `#include <stdexcept>`
- `csrc/common/exception.h`: added `#include <stdexcept>` (the `std::runtime_error` branch is compiled under the MindSpore build)
- `csrc/mindspore/framework_hal.cpp` and `csrc/mindspore/mem_kernels.cpp`: same fix, since they use `std::runtime_error` too

Together with 3091f21 (mem_alloc.cpp), all files using `std::runtime_error` now include `<stdexcept>` directly. Ready to merge whenever you are.

chloroethylene:
fix lint plz

pjh456:
Fixed the clang-format failure in 1736ffc. The issue was include
ordering: `<stdexcept>` in `csrc/pybind.cpp` had to come after
`<pybind11/stl.h>` to satisfy clang-format's lexicographic sort.

Verified all 5 touched files are now clean under `clang-format --style=LLVM`.

CI should pass now. Let me know if anything else is needed.

matthewygf (approved):
LGTM ! Thanks for the contribution.

PR #291, opened 2026-09-07
[PD][P2P] Enable combined PD/P2P operation and token-aware proxy routing on Ascend

shengyan-chen:
## Overview

This PR enables PD and P2P to work together on the Ascend sender path and adds a token-aware proxy for disaggregated prefill/decode serving.

It combines the lifecycle improvements from #273 with proxy routing and request handling, targeting `dsv4_support_045` while preserving its existing multi-group KV support. The changes are not specific to a single model.

## Impact

The backend changes address mixed PD/P2P initialization, incorrect key/object pairing after skipped allocations, shared receiver entry ownership, and preemption metadata propagation(for vllm 0.23.0).

The proxy adds token-aware load distribution across prefill and decode instances, transfer-mode-aware admission control, and separate handling for Completion and native Chat requests.

### Configuration and sender-side correctness
- Allow `enable_pd=true` and `enable_p2p=true` for `pd_role="sender"`, with explicit validation and upstream LMCache 0.4.5 checks preserved.
- Provision LocalCPUBackend for mixed senders even when local CPU caching is disabled, and report missing capacity explicitly.
- Track allocated keys alongside copied objects, preserving key/object alignment when existing keys are skipped.
- Retain multi-group shape/dtype allocation and device-independent stream handling.
- Clamp producer store skip to tokens already transferred to the decoder.

### Receiver lifecycle and preemption
- Track receiver entries and request ownership through lookup/retrieve contexts.
- Use request-local delay-pull proxy clones so consuming one request's proxy does not invalidate another request's hit.
- Release request leases during lookup unpin and retrieve cleanup; defer PD Done signaling until the final active request lease is released.
- Remove both receiver indexes during partial push allocation rollback without double-releasing objects.
- Carry scheduler preemption IDs alongside multi-group connector metadata, retaining legacy set payload compatibility and existing cleanup/drain behavior.
- Include #274 as a prerequisite and preserve its proxy refcount/pin handling.

### Token-aware disaggregated proxy
- Introduce the upstream-derived proxy and extend it with prompt-token-aware prefill/decode routing.
- Track per-instance request/token load and expose routing decisions and service timing metrics.
- Apply best-effort weighted buffer admission in push and eager-pull modes; disable buffer-size admission for delay-pull.
- For Completion requests, reuse the prefiller's first output token, adjust the remaining decode budget, and handle single-token requests without invoking the decoder.
- For Chat requests, render prompt tokens through vLLM, use an internal prefill request, and retain native Chat decoding with the full client-visible output budget.
- Preserve upstream HTTP error details and release proxy accounting resources on tested failure and cancellation paths.

## Tests

Added and adjusted regression coverage for:
- Mixed sender configuration, upstream validation, and LocalCPUBackend initialization.
- Key/object alignment, multi-group byte copying, and producer skip clamping.
- Shared receiver leases, request-local proxy isolation, final-lease Done signaling, and allocation rollback.
- Multi-group metadata preservation, preemption payload compatibility, and retrieve context cleanup.
- Proxy request construction, routing/accounting, transfer-mode admission, native Chat responses, and failure/cancellation cleanup.

## Scope note

This PR retains the existing multi-group implementation rather than introducing a new model-specific KV layout.

Delay-pull is the primary deployment mode, with identically configured prefill instances and a validated homogeneous P/D topology. Generalized multi-group buffer capacity estimation and heterogeneous topology support are deferred; proxy admission remains best-effort.

Additional handoff-ID propagation and protection for the interval before decoder request leases are established remain follow-up work.
Validation:
- Pre-commit checks passed for the migrated changes.
- Proxy tests passed: 24 cases, run independently of the NPU test bootstrap.
- The original implementation was end-to-end tested before migration. Full backend/NPU and end-to-end validation on this target branch remains pending.

## Scope note

This PR retains the existing multi-group implementation rather than introducing a new model-specific KV layout.

Delay-pull is the primary pd deployment mode, with identically configured prefill instances and a validated homogeneous P/D topology. Generalized multi-group buffer capacity estimation(in proxy) and heterogeneous topology support are deferred; proxy admission remains best-effort.

Additional handoff-ID propagation and protection for the interval before decoder request leases are established remain follow-up work.

### DIFF
--- a/examples/disagg_prefill/disagg_proxy_request.py
+++ b/examples/disagg_prefill/disagg_proxy_request.py
@@ -1,0 +1,111 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Standard
+from typing import Any
+
+DEFAULT_COMPLETION_MAX_TOKENS = 16
+
+
+class UpstreamServiceError(Exception):
+    """An HTTP error response returned by a proxied vLLM service."""
+
+    def __init__(
+        self,
+        *,
+        status_code: int,
+        url: str,
+        body: bytes,
+        content_type: str | None,
+    ) -> None:
+        self.status_code = status_code
+        self.url = url
+        self.body = body
+        self.content_type = content_type
+        self.body_text = body.decode("utf-8", errors="replace")
+        super().__init__(
+            f"Upstream service returned {status_code} for {url}: {self.body_text}"
+        )
+
+
+def upstream_service_error_from_response(response: Any) -> UpstreamServiceError:
+    """Capture an upstream HTTP response before its details are discarded."""
+    return UpstreamServiceError(
+        status_code=response.status_code,
+        url=str(response.request.url),
+        body=response.content,
+        content_type=response.headers.get("content-type"),
+    )
+
+
+def normalize_chat_request(request_data: dict[str, Any]) -> dict[str, Any]:
+    """Normalize semantically empty tool configuration for vLLM chat APIs."""
+    normalized = request_data.copy()
+    if normalized.get("tools") == [] and normalized.get("tool_choice") in (
+        None,
+        "none",
+    ):
+        normalized.pop("tools")
+    return normalized
+
+
+def parse_chat_render_output(
+    render_output: dict[str, Any],
+) -> tuple[list[int], int]:
+    """Extract rendered prompt tokens and the effective output budget."""
+    return (
+        list(render_output["token_ids"]),
+        render_output["sampling_params"]["max_tokens"],
+    )
+
+
+def build_chat_phase_requests(
+    request_data: dict[str, Any],
+    prompt_token_ids: list[int],
+) -> tuple[dict[str, Any], dict[str, Any]]:
+    """Build an internal prefill request and a native Chat decode request."""
+    prefill_request = request_data.copy()
+    prefill_request.pop("max_completion_tokens", None)
+    prefill_request.pop("stream_options", None)
+    prefill_request["prompt"] = list(prompt_token_ids)
+    prefill_request["max_tokens"] = 1
+    prefill_request["stream"] = False
+
+    decode_request = request_data.copy()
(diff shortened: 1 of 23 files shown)


matthewygf:
I found two deterministic lifecycle blockers in the new proxy server at `6137cf6`:

1. **Shutdown can hang forever when no notification arrives.** `lifespan()` sets `run_proxy = False` and awaits `zmq_task`, but that task may already be suspended in `await socket.recv()`. The socket has no receive timeout/poll timeout and is not cancelled or closed by the shutdown path, so the loop condition is never re-evaluated. I reproduced this with a blocking fake `recv()`: lifespan teardown timed out instead of completing. The `zmq.Again` branch cannot make progress with the default blocking receive.

   Please cancel-and-await the task during teardown (handling `CancelledError`) and close the socket in a `finally`, or use bounded polling/receive timeouts. Add a regression that enters/exits lifespan without delivering a ZMQ message.

2. **Cancelling a non-streaming chat request leaks decoder accounting and PD buffer permits.** After acquiring both resources, the non-stream path awaits KV readiness and the decoder request. Its only outer cleanup is `except Exception`, but `asyncio.CancelledError` is a `BaseException` on supported Python versions. Cancellation at either await bypasses both `release_decoder()` and `release_pd_buffer_slots()`. A focused reproduction cancelling from `wait_decode_kv_ready()` observed one prefiller release but zero decoder and slot releases. Repeated disconnects will permanently inflate decoder load and can exhaust the weighted semaphore.

   Please put decoder/slot cleanup in a cancellation-safe `finally` (or explicitly handle `BaseException`) and add a non-streaming cancellation regression. The existing cancellation test exercises only the completion streaming generator, whose `finally` already releases these resources.

The 24 focused proxy tests currently pass, but neither sequence is covered. Changed-file syntax compilation and `git diff --check` pass.

PR #292, opened 2026-09-08
[Qwen3.5/Qwen3.8][Hybrid Cache] Add GDN state checkpoint storage and reuse on Ascend

shengyan-chen:
# [Qwen3.5][Hybrid Cache] Add GDN state checkpoint storage and reuse on Ascend

## Overview

This PR adds hybrid caching for Qwen models with Full Attention and Gated DeltaNet (GDN) layers on Ascend. It combines chunk-based Attention KV caching with independent recurrent-state checkpoints, and supports prefill checkpoint reuse followed by Qwen MTP decoding.

Attention KV describes a token interval, while a GDN checkpoint represents the state after a complete prefix, `S(R)`. An Attention hit alone is insufficient for hybrid reuse. The connector selects a common restore boundary supported by Attention coverage and every required GDN group and TP rank.

The implementation covers registration, checkpoint allocation, native state transfer, sparse lookup, and CPU/disk storage integration with LMCache 0.4.5. Deployment examples and real-model validation cover Qwen3.5-122B-A10B and Qwen3.8-27B with TP8 and MTP3.

### Group registration and checkpoint layout

- Classify Attention and GDN layers from cache specs, validate runtime tensors, and select an unambiguous Full Attention primary group.
- Retain supported MTP Attention layers in the existing Attention transfer path, preserving their tensors and scheduler mappings.
- Keep each GDN layer's conv and SSM planes in the same scheduler group.
- Allocate one managed `MemoryFormat.BINARY` object per group checkpoint using the existing CPU allocator and reference lifecycle.
- Expose plane-major, mixed-dtype `buffer.planes` views, preserving native shapes, layer order, offsets, and layout compatibility metadata.

### Native state transfer

- Add a GDN state copy operator for store and restore, using plane/layer pointer ordering and contiguous runtime planes.
- Copy the complete conv plane, including MTP-reserved capacity, and one SSM state from the selected physical block per layer. Shapes come from the actual spec and tensors.
- Use `DataCopyPad` with exact byte lengths for small payloads and non-aligned tails, including adjacent layer slices.
- Complete state transfer before publishing the captured buffer or returning from restore.

### Sparse checkpoint save, lookup, and restore

- Reuse LMCache's chained prefix hashes with separate state keys identifying the group and checkpoint boundary. Preserve layout compatibility in object metadata.
- Keep the execution endpoint, Attention save endpoint, and state checkpoint boundary distinct. Attention clipping does not change which state snapshot is captured.
- Save state only at a real, eligible aligned execution endpoint; an Attention chunk does not imply that a state checkpoint exists.
- Select one restore boundary across all required GDN groups and TP ranks, with Attention coverage for the missing interval. Re-query sparse candidates until ranks agree.
- Apply the full-hit lookup limit before choosing a checkpoint, keeping the returned token boundary consistent with the restored state.
- Retain selected state buffers and Attention pins through consumption or cleanup.
- Restore Attention and GDN state using the runner's block mappings before its subsequent state movement and model execution.
- Raise on missing or incompatible required state, incomplete Attention restoration, or transfer failure. Log request/rank/group/boundary context and suppress subsequent saves for locally failed requests.

### MTP prefill reuse

- Allow the supported Qwen target/draft configuration while retaining `align` state management and synchronous scheduling and transfer.
- Determine prefill/decode phase from request progress before building save metadata. This handles both a final one-token prefill and multi-token MTP decode steps.
- In MTP mode, save Attention and GDN state only for eligible prefill steps without scheduled draft tokens. Reject `save_decode_cache=true`; decode steps cannot publish new external checkpoints.
- Track relocated speculative state blocks, prefer complete allocation tables when available, and clear invalid old positions when applying incremental updates.
- Use the runner's existing post-draft finalization for saves. Restore the selected checkpoint before upstream state movement; accepted-token handling remains with vLLM/vLLM-Ascend.
- Reuse the existing common-boundary lookup without adding an MTP-specific hash or an extra-block lookup rollback rule.
- Permit eager and graph execution configurations. Model-specific validated settings and the known Qwen3.8 graph limitation are listed below.

### Storage, logging, and deployment

- Preserve state metadata and mixed-dtype segment descriptions through CPU and disk reads. Publish disk entries after complete writes and validate restored payload sizes.
- Manage allocation references, lookup pins, pending writes, and disk capacity through the covered lifecycle and failure paths.
- Align hybrid lookup logging with existing cache logs. Add state store/restore logs with request ID, rank, boundary, groups, payload size, elapsed time, and throughput.
- Add launch scripts and LMCache YAML examples under `docs/qwen`, with INFO logging, user-editable path placeholders, and optional disk settings commented out.

## LMCache storage format

The following diagram shows one TP rank. Group IDs and layer names are illustrative; different GDN groups use their own runtime block mappings.

```mermaid
flowchart LR
    subgraph Runtime["vLLM runtime on NPU"]
        Mapping["GDN group g: selected block b at boundary R"]
        L0["Layer L0: conv[b] and SSM[b]"]
        L1["Layer L1: conv[b] and SSM[b]"]
        Ln["Layer Ln: conv[b] and SSM[b]"]
        KV["Attention layers, including MTP when enabled:<br/>K/V blocks and token slots"]
        Mapping --> L0
        Mapping --> L1
        Mapping --> Ln
    end

    subgraph Cache["LMCache objects in local CPU / disk storage"]
        State["One group checkpoint S(R)<br/>conv: L0 | L1 | ... | Ln<br/>SSM: L0 | L1 | ... | Ln"]
        Meta["Separate key and metadata<br/>prefix identity, group, R<br/>layout version, shapes, dtypes, offsets"]
        Attention["Attention payloads<br/>existing token-chunk organization"]
        Meta --- State
    end

    L0 -->|store| State
    L1 -->|store| State
    Ln -->|store| State
    KV -->|store| Attention
```

Load reverses these transfers. A runtime plane has shape `(num_blocks, *state_shape)`; a checkpoint plane has shape `(group_layer_count, *state_shape)`. Only the selected state slot from each layer is copied.

```mermaid
flowchart TB
    subgraph Namespace["Existing model / TP-rank key namespace"]
        Prefix["Chained prefix hash H_R<br/>Complete prefix ending at R"]
        AttentionObject["Attention object, if present<br/>KV for the chunk ending at R"]
        State0["GDN group 0 object, if present<br/>S_0(R)"]
        State1["GDN group 1 object, if present<br/>S_1(R)"]
        State2["GDN group 2 object, if present<br/>S_2(R)"]
        Prefix -.->|prefix identity| AttentionObject
        Prefix -.->|prefix identity + group + R| State0
        Prefix -.->|prefix identity + group + R| State1
        Prefix -.->|prefix identity + group + R| State2
    end
```

The dotted edges show logical key associations, not a shared allocation or an on-disk directory tree. Each object can be present or absent independently.

```mermaid
flowchart TB
    subgraph Allocation["One GDN group checkpoint: one managed allocation"]
        ConvRegion["offset_conv<br/>L0 conv | L1 conv | ... | Ln conv"]
        SSMRegion["offset_ssm<br/>L0 SSM | L1 SSM | ... | Ln SSM"]
        ConvRegion -->|increasing byte offset| SSMRegion
    end

    ConvView["buffer.planes[0]<br/>shape: (group_layer_count, *conv_shape)<br/>native conv dtype"]
    SSMView["buffer.planes[1]<br/>shape: (group_layer_count, *ssm_shape)<br/>native SSM dtype"]
    ConvView -.->|typed view, no copy| ConvRegion
    SSMView -.->|typed view, no copy| SSMRegion
```

These are independently stored objects, not one combined Attention/state allocation. A chunk key identifies the full prefix through chained hashing, but its Attention payload contains only that chunk's KV. Restoring at `R` also requires the preceding Attention chunks needed for the missing interval.

Checkpoint size depends on group layout, not prefix length. The two plane views share one allocation without additional copies. Identity and layout metadata are separate from the raw payload, rather than an inline binary header. Disk payloads preserve the allocation's byte layout and are interpreted using backend metadata; restart-time disk-index recovery is outside this PR's scope.

## Validation

Core regression coverage under `tests/v1` includes:

- Spec classification, primary selection, layout validation, mixed-dtype allocation, and buffer lifecycle.
- Native state round trips, invalid inputs, tile boundaries, small payloads, non-aligned tails, and multi-layer copies.
- Sparse cross-rank lookup, independent checkpoint storage, metadata compatibility, and pin/reference cleanup.
- Execution boundaries, block mappings, partial Attention coverage, and injected load failures.
- MTP registration, execution-mode configuration, prefill-only save gating, speculative block relocation, and complete-table replacement.
- CPU/disk object lifecycle and affected adapter/storage behavior.

The native extension built successfully on the 910B3 server with CANN 9.1.0. The previously reported targeted transfer/integration run passed 84 tests; that count predates the later MTP additions and is not a full-suite count for this revision. Targeted MTP unit testing and real-model testing were subsequently performed on the NPU server. An opt-in real-model regression test is also provided under `tests/v1`; the broader acceptance datasets and audit tools remain local validation artifacts.

### real-model acceptance

Both models were tested with BF16, TP8, MTP3, synchronous hybrid caching, and the local CPU tier:

| Model | Target execution | Draft execution | LMCache chunk size | Answer checks | All-rank restores confirmed |
| --- | --- | --- | --- | --- | --- |
| Qwen3.5-122B-A10B | `FULL_DECODE_ONLY` | Eager | 1024 | 30/30 | 15/15 |
| Qwen3.8-27B | Eager | Eager | 768 | 30/30 | 15/15 |

For each model, acceptance covered:

- Twelve prompt cases from 256 tokens to approximately 32K tokens, including chunk boundaries and long prefixes.
- Cold requests, serial external-cache reuse, and six concurrent reuse requests with a peak of four overlapping client requests.
- Exact expected-answer checks on 30 formal responses. One-token donor requests populated cache entries and were excluded from answer checks.
- Restores leaving 1, 3, 4, and 5 prefill tokens, including the MTP3 `num_speculative_tokens + 1` boundary.
- Clearing vLLM's local prefix cache while retaining LMCache entries, and auditing restore boundaries and all eight TP ranks to confirm external reuse.

Earlier Qwen3.5 testing also exercised disk lookup and restore. The latest two-model acceptance above used CPU storage; it does not establish disk coverage for both final configurations.

These are functional acceptance results, not a general accuracy or performance benchmark. They do not establish 128K context coverage, 32-client capacity, or exact internal state equality. A separate Qwen3.5 reproduction run confirmed graph dispatch for the four-token restore-tail cases; the latest acceptance summary alone does not prove graph execution on every step.

## Supported scope and limitations

- Text-only hybrid inference with `align` state management, sparse checkpoints, synchronous state capture/restore, and local CPU/disk storage. Disk persistence retains the backend's asynchronous write lifecycle.
- MTP support is limited to prefill checkpoint reuse followed by decode. No MTP decode checkpoint saving or speculative accepted-state persistence is added.
- Supported MTP registration uses the Qwen3.5-family target/draft model identities and compatible Full Attention grouping. This does not enable arbitrary speculative methods or separate Eagle groups.
- **Qwen3.8-27B currently uses eager execution.** Earlier graph-mode tests produced truncated outputs after reuse in short-prefill cases, including a four-token tail with MTP3. The corresponding eager cases passed. The graph issue remains unresolved; this PR does not claim to fix it. The tested Qwen3.5 graph cases did not reproduce that failure.
- Pipeline parallelism, multimodal requests, blending, layerwise/asynchronous hybrid transfer, PD/P2P, remote storage, and restart-time disk-index recovery remain outside this scope.
- This PR does not modify vLLM or vLLM-Ascend. Required hybrid load failures are propagated rather than returning `false` and allowing forward execution with incomplete state. Coordinated all-rank rollback/recompute is not implemented, so a load failure may terminate serving instead of transparently recomputing the affected request.

## Deployment examples and integration dependency

- Qwen3.5: `docs/qwen/start_qwen35_lmcache.sh` and `docs/qwen/lmcache_qwen35.yaml`.
- Qwen3.8: `docs/qwen/start_qwen38_lmcache.sh` and `docs/qwen/lmcache_qwen38.yaml`.

The examples use TP8 and MTP3, with model-specific execution modes and chunk sizes. Replace `/xxx/` paths before use. Disk configuration is optional and commented out by default.

The `third_party/kvcache-ops` submodule points to `c3db0eb71c33957941a0b0cbfbc002aea5607796` and currently uses the contributor-hosted `https://gitcode.com/syan0o0/kvcache-ops.git` source. Upstream integration of the operator and the final submodule source still need to be resolved before merge.

## kvcache-ops pr
https://gitcode.com/openeuler/kvcache-ops/pull/27

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "third_party/kvcache-ops"]
 	path = third_party/kvcache-ops
-	url = https://atomgit.com/openeuler/kvcache-ops.git
+	url = https://gitcode.com/syan0o0/kvcache-ops.git
 
 [submodule "third_party/hcomm"]
 	path = third_party/hcomm
--- a/csrc/hixl/CMakeLists.txt
+++ b/csrc/hixl/CMakeLists.txt
@@ -15,6 +15,7 @@ target_include_directories(hixl_npu_comms PRIVATE
     ${ASCEND_CANN_PACKAGE_PATH}/include/experiment
     ${ASCEND_CANN_PACKAGE_PATH}/include/experiment/runtime
     ${ASCEND_CANN_PACKAGE_PATH}/${ARCH_SUBDIR}/include
+    ${ASCEND_CANN_PACKAGE_PATH}/${ARCH_SUBDIR}/pkg_inc
     ${ASCEND_CANN_PACKAGE_PATH}/${ARCH_SUBDIR}/pkg_inc/runtime
     ${ASCEND_CANN_PACKAGE_PATH}/${ARCH_SUBDIR}/include/experiment/msprof
 )
--- a/csrc/mem_kernels.cpp
+++ b/csrc/mem_kernels.cpp
@@ -3,12 +3,142 @@
 #include "utils.h"
 #include <ATen/ATen.h>
 #include <Python.h>
+#include <array>
+#include <cstdint>
+#include <limits>
 #include <pybind11/pybind11.h>
 #include <torch_npu/csrc/core/npu/NPUStream.h>
 #include <torch_npu/csrc/framework/OpCommand.h>
 #include <torch_npu/csrc/npu/Module.h>
 
 namespace py = pybind11;
+
+namespace {
+
+torch::Tensor build_gdn_state_ptr_tensor_on_device(
+    const std::vector<torch::Tensor> &state_tensors, int64_t plane,
+    int64_t num_layers, const torch::Device &runtime_device) {
+  auto state_ptrs_cpu = torch::empty(
+      {num_layers},
+      torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU));
+  auto *state_ptrs_cpu_data = state_ptrs_cpu.data_ptr<int64_t>();
+  for (int64_t layer = 0; layer < num_layers; ++layer) {
+    state_ptrs_cpu_data[layer] =
+        static_cast<int64_t>(reinterpret_cast<uintptr_t>(
+            state_tensors[plane * num_layers + layer].data_ptr()));
+  }
+  return state_ptrs_cpu.to(runtime_device);
+}
+
+} // namespace
+
+void multi_layer_gdn_state_transfer(std::vector<torch::Tensor> &memory_tensors,
+                                    std::vector<torch::Tensor> &state_tensors,
+                                    const int64_t block_id,
+                                    const bool direction) {
+  TORCH_CHECK(memory_tensors.size() == 2,
+              "GDN transfer expects exactly 2 memory tensors.");
+  TORCH_CHECK(!state_tensors.empty() && state_tensors.size() % 2 == 0,
+              "GDN transfer expects two non-empty runtime tensor families.");
+  TORCH_CHECK(block_id >= 0, "GDN block_id must be non-negative.");
+  const auto num_layers = static_cast<int64_t>(state_tensors.size() / 2);
+  TORCH_CHECK(num_layers <= std::numeric_limits<int32_t>::max(),
+              "GDN num_layers exceeds the kernel limit.");
+  TORCH_CHECK(state_tensors[0].defined(),
+              "GDN runtime tensor must be defined.");
+  const auto runtime_device = state_tensors[0].device();
+  TORCH_CHECK(runtime_device.is_privateuseone() && runtime_device.index() >= 0,
+              "GDN runtime tensors must be on an actual NPU device.");
+  const c10::OptionalDeviceGuard device_guard(runtime_device);
+
+  std::array<uint8_t *, 2> memory_ptrs;
+  std::array<GDNStateTransferConfig, 2> configs;
+  // Validate both planes before submitting either copy. In particular, a bad
+  // SSM mapping must not leave an otherwise valid conv payload half-written.
+  for (int64_t plane = 0; plane < 2; ++plane) {
+    auto &memory = memory_tensors[plane];
+    TORCH_CHECK(memory.defined(), "GDN memory tensor must be defined.");
+    TORCH_CHECK(memory.dim() >= 2 && memory.size(0) == num_layers,
+                "GDN memory shape must be [num_layers, *state_shape].");
+    TORCH_CHECK(memory.is_contiguous() && memory.numel() > 0,
+                "GDN memory tensor must be non-empty and contiguous.");
+    TORCH_CHECK(
+        memory.device().is_cpu() || memory.device() == runtime_device,
+        "GDN memory must be registered CPU memory or on the runtime NPU.");
+    const auto dtype = memory.scalar_type();
+    TORCH_CHECK(dtype == at::ScalarType::Float ||
+                    dtype == at::ScalarType::BFloat16,
+                "GDN transfer supports only FP32 and BF16, got ", dtype, ".");
+    TORCH_CHECK(
+        kvcache_ops::gdn_state_transfer_supports_dtype(
+            vllm_ascend::get_dtype_from_torch(dtype)),
+        "GDN dtype is not supported by this kernel build; BF16 requires "
+        "ASCEND_AICORE_ARCH >= 220.");
+    for (int64_t layer = 0; layer < num_layers; ++layer) {
+      const auto &state = state_tensors[plane * num_layers + layer];
+      TORCH_CHECK(state.defined(), "GDN runtime tensor must be defined.");
+      TORCH_CHECK(state.device() == runtime_device,
+                  "All GDN runtime tensors must be on the same NPU device.");
+      TORCH_CHECK(state.is_contiguous(),
+                  "GDN runtime tensor must be contiguous.");
+      TORCH_CHECK(state.scalar_type() == dtype,
+                  "GDN runtime and memory dtype mismatch at plane ", plane,
+                  ", layer ", layer, ".");
+      TORCH_CHECK(state.dim() == memory.dim(), "GDN runtime rank mismatch.");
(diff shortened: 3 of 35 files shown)


matthewygf:
I found a **high-severity silent-correctness blocker** in the hybrid restore path at `b45e961`.

`get_num_new_matched_tokens()` returns `R - C`, so vLLM schedules the request as though tokens `[C, R)` are already computed. At worker load time, however, `_load_hybrid_request()` catches every detected `StateLoadError` (expired/missing selection, missing buffer or metadata, stale mapping, incomplete Attention coverage), records the request ID, and returns `False`. `start_load_kv()` ignores that return value and returns normally. Nothing resets the scheduler's computed-token count or marks blocks invalid; the tests explicitly assert that these failures leave `_invalid_block_ids` empty.

A deterministic sequence is:

1. Scheduler lookup selects `R > C` and returns `R - C` external tokens.
2. The worker's selection is unavailable by `start_load_kv()` (for example, it expires between lookup and execution), so `get_state_lookup()` yields no usable selection.
3. `_load_hybrid_request()` records failure and returns `False`.
4. `start_load_kv()` discards `False` and forward execution continues while `[C, R)` remains skipped, with recurrent state not restored. In TP, ranks can also diverge if only one rank hits this failure.

The later `_failed_state_loads` checks only suppress publishing the corrupted suffix; they do not repair the current forward pass. The ordinary multi-group path handles incomplete KV retrieval through `record_failed_blocks()` and `_invalid_block_ids`, but this hybrid early return bypasses that recovery.

Please make a detected state-load failure impossible to return from `start_load_kv()` as success. A safe minimum for this PR is to propagate/fail fast so the request or batch aborts; transparent recomputation needs an all-rank scheduler/worker recovery protocol. Add a regression at the public `start_load_kv()` boundary showing that a missing/expired selection (and incomplete Attention coverage) cannot return normally after the scheduler counted external tokens.

Issue #28, opened 2025-10-21
[Bug] Adding extra keys to the python standard hashfunc produce inconsistent hash across workers


matthewygf:
The bug exists in LMCache-Ascend Docker built environment using v0.3.7-dev branch.

- When extra_keys is None, the hashed produced across workers give inconsistent hash given the same tokens, if removing the extra_keys, the hashes across workers are consistent.

See the related issue: https://github.com/python/cpython/issues/99540

I have verified that if we turned ofof randomized_va in the kernel we get the consistent hash

```python
...
 return self.hash_func((prefix_hash, tokens_tuple, extra_keys)) # problem matic
```

matthewygf:
Introduced from #16

PR #297, opened 2026-09-17
[MP] Support MP mode.

marcobarlo:
## Summary

Lands **multiprocess (MP) mode** on official LMCache-Ascend `main`: a native NPU `multi_layer_block_kv_transfer` that matches upstream LMCache’s CUDA c_ops surface, plus the plugin glue to bind it.

This is the NPU half of [LMCache#5138](https://github.com/LMCache/LMCache/pull/5138). That PR adds `EngineKVFormat.NL_X_NP_X_NB_BS_ONE_HS` (fmt **17**) so vLLM-Ascend MLA/DSA plane tuples — `(latent, rope)`, `(latent, rope, dsa)`, or DSv4 `(int8 latent, fp16 scale)` — are detected and stored as **one packed object** (`kv_size == 1`). Grouping, format detection, and the Python `torch_ops` fallback stay in LMCache. This plugin only implements the **native copy** for those layouts (plus fmt **16** separate K/V and fmt **13** NH_CS).

### Architectural choices

- **Same C API as CUDA.** `PageBufferShapeDesc` is taken **by value**; `EngineKVFormat` / `TransferDirection` numeric values match CUDA pybind so `NpuDeviceOps.bind_native` can attach the same symbols. Packed geometry is `plane_slot_bytes` + `plane_block_stride_bytes` (from #5138), not model-named kernel branches.
- **Split of responsibility.** Host grouping, object keys, IPC, and `lmcache_driven` orchestration live in LMCache-MP. The plugin supplies the AscendC kernel, stream recorders, `lmcache_memcpy_async`, and a thin `_patch_ops` wrap (fmts 13/16/17 → native; everything else → `torch_ops`).
- **`lmcache_driven` on NPU.** Worker exports KV IPC handles once; the **server** runs `multi_layer_block_kv_transfer` on imported views, then `lmcache_memcpy_async` into pinned L1. Plugin IPC shims (`custom_types`, `server`, `tokens_hash`) are deleted — LMCache-MP already owns wrappers / events via `torch_npu`.
- **NPU launch constraints.** Do **not** drop the GIL around `OpCommand.Run()` (torch_npu posts to a Python task queue; CUDA does not). Lazy L1 pool uses `aclrtMallocHost` instead of post-hoc `aclrtHostRegister`.
- **Partial import.** `NpuDeviceOps.ensure_native` imports this plugin in the middle of `import lmcache`. `_seed_partial_lmcache_device_ops` installs a lazy `lmcache.device_ops` proxy so core modules do not `ImportError` and silently fall back to the ~8× slower torch baseline.

```mermaid
flowchart LR
  vLLM[vLLM-Ascend worker<br/>paged KV planes] --> Conn[LMCache MP connector]
  Conn --> Det["Format + groups<br/>LMCache #5138<br/>fmt 17 packed / 16 K-V / 13 NH_CS"]
  Det --> Srv[LMCache server]
  Srv --> Wrap["_patch_ops wrap"]
  Wrap -->|13 / 16 / 17| Native["lmcache_ascend.c_ops<br/>multi_layer_block_kv_transfer"]
  Wrap -->|other fmts| Torch[torch_ops fallback]
  Native --> Ker[AscendC byte mover<br/>plane_slot_bytes + strides]
  Ker -->|D2H| Pin[pinned MemoryObj]
  Ker -->|H2D| Paged[worker paged KV]
```

```mermaid
sequenceDiagram
  participant W as vLLM worker
  participant S as LMCache server
  participant K as NPU c_ops (this PR)
  W->>S: register KV IPC handles + engine_group_infos
  Note over S: #5138 classifies planes as fmt 17
  W->>S: STORE / RETRIEVE (block_ids + stream event)
  S->>K: multi_layer_block_kv_transfer(desc by value)
  K-->>S: staging / paged KV
  S->>K: lmcache_memcpy_async (staging ↔ pinned L1)
```

## Commits

1. **Native transfer + recorders** — AscendC `multi_layer_block_mem_kernels`, `mp_mem_kernels` host prepare/launch, event/completion recorders, pybind. Exports `EngineKVFormat` (including fmt 17 from #5138), `PageBufferShapeDesc` packed-plane fields, plan types (`KernelGroupSpec` / `StagingCopy` / `execute_object_group_transfer`), and `lmcache_memcpy_async`. CMake copies `new_kernels/` into the kvcache-ops GLOB. 
The kernel uses a single DataCopy when memory is contiguous for all tokens (e.g., K, V), and it uses DataCopyPad with striding to/from GM when the Memobj is arranged token-major with different planes for the same token packed contiguously on the same row. This choice is borrowed from the SP mode multi-plane kernel and mirrors upstream CUDA. E.g., MLA with (latent, rope) where latent and rope have different dimensions, and DSv4 (latent, scale), where scale is only 2 Bytes. Padding is used when needed to respect UB 32-Bytes alignment constraints.

2. **Plugin MP glue** — `_patch_ops` wraps the native op for fmts 13/16/17 (nested layer lists → int64 pointer table; keep GIL). `shape_desc.py` attaches `plane_slot_bytes` / strides; `logical_block_size` reports the compressed token span without mutating `spec.block_size`. Deletes NPU IPC / hash shims now owned by LMCache-MP.

3. **LMCache 0.5 lockstep** — Not the kernel. Mechanical import path for LMCache-MP 0.5: `memory_allocators.*`, `attempt_permute_to_contiguous_view`, `MemoryObj.get_shapes()` / `get_dtypes()`, `GPUKVFormat = 8`. `_seed_partial_lmcache_device_ops` + `aclrtMallocHost` lazy pool so activation and the cache-server pin succeed. Required to *import*; CUDA #5138 still needs this plugin for a native NPU copy.

4. **MP tests** — Direct kernel roundtrip (`test_mp_block_transfer.py`), LMCache-driven e2e (`ObjectKey` + `closing()`, fmt 17 only), NPU `test_custom_types`. `conftest` collects those MP files instead of skipping all `v1/multiprocess/test_*.py`.

5. **Test collapse** — One `_build(layout, …)` for packed MLA, NH_CS, and separate K/V. Same contracts (roundtrip × host, native vs `torch_ops`, plan vs direct, `TORCH_CHECK` → Python `RuntimeError`); ~1090 → ~720 lines.

## Architecture vs upstream LMCache

| Upstream CUDA / #5138 | This PR (NPU) |
|---|---|
| Detector + `NL_X_NP_X_NB_BS_ONE_HS` spec ([#5138](https://github.com/LMCache/LMCache/pull/5138)) | Consumes that fmt; does not re-detect |
| `c_ops.multi_layer_block_kv_transfer(PageBufferShapeDesc by value)` | Same signature / enum values (fmt **16** / **17** / **13**) |
| Packed MLA via `plane_slot_bytes` / `plane_block_stride_bytes` | Same; kernel is a byte mover |
| `execute_object_group_transfer` + plan types | Same; **no GIL release** around `OpCommand.Run()` |
| CUDA IPC in core | Plugin IPC deleted; `torch_npu` handles in LMCache-MP |
| `bind_native` mid-`import lmcache` | `_seed_partial_lmcache_device_ops` |

## Tested with

| Component | Version |
|---|---|
| **LMCache** | LMCache-MP `0.5.5rc6.dev72` (`718f6c03`, `nightly-musa-44-g718f6c03`) |
| **vLLM** | `0.23.0` (`0fc695f`, tag `v0.23.0`) |
| **vLLM-Ascend** | `0.23.0` (`5cb98ca`, image `quay.io/ascend/vllm-ascend:v0.23.0`) |
| **torch-npu** | `2.10.0.post4` |
| **Hardware** | 8× Ascend 910B3, container `vllm-ascend-dsv4-lmcache` |

**Regression Tests against https://github.com/marcobarlo/LMCache-Ascend/commits/MP_dsv4_3_split:** `test_mp_block_transfer.py` 37 passed; nh_cs + LMCache-driven e2e `[0]` passed. **Serving:** DSv4 Flash w8a8 TP=8, short and long prompt store/load OK. Tested against https://github.com/marcobarlo/LMCache/commits/MP_dsv4_3_split/

**Note:** official plugin tag is still `LMCACHE_UPSTREAM_TAG = v0.4.4`. Commit 3 is required for LMCache **0.5** (`memory_allocators`, permute rename). CI against pip `lmcache==0.4.4` will fail those imports.


matthewygf:
I found two merge-blocking integration gaps at the current head (`21e0c7f`). They are independent of the acknowledged v0.4.4 import mismatch.

1. **High — the advertised NPU MP cache context cannot be constructed from the referenced upstream PR.** The new end-to-end test calls [`NpuDeviceSpec().create_cache_context(...)`](https://github.com/LMCache/LMCache-Ascend/blob/21e0c7f10fc0178eea0b731cf11490ac12e417bc/tests/v1/multiprocess/test_npu_lmcache_driven_e2e.py#L216-L226), but LMCache#5138 head `704d183` defines `NpuDeviceSpec` without a `create_cache_context` override; the inherited base method raises `NotImplementedError`. This plugin branch does not install an override either. The hardware result in the description was run against separate `MP_dsv4_3_split` forks, so it does not validate the dependency graph named by this PR. Please land/reference the missing core factory implementation (or provide it here) and run the e2e against those exact commits.

2. **High — fmt 17 production descriptors never receive the plane geometry required by the native kernel.** [`prepare_group` rejects fmt 17 when `num_planes == 0`](https://github.com/LMCache/LMCache-Ascend/blob/21e0c7f10fc0178eea0b731cf11490ac12e417bc/csrc/mp_mem_kernels.cpp#L115-L126). This branch adds [`attach_tuple_planes` and `attach_tuple_block_strides`](https://github.com/LMCache/LMCache-Ascend/blob/21e0c7f10fc0178eea0b731cf11490ac12e417bc/lmcache_ascend/v1/shape_desc.py#L18-L53), but no production code calls either helper. LMCache#5138's `make_page_buffer_shape_desc` populates only the legacy scalar fields and returns, so an MLA/DSA registration reaches the native transfer with the sentinel unchanged and fails before launch. The direct kernel tests mask this by calling the helpers manually. Please wire the real per-plane tensors into descriptor creation and assert in the LMCache-driven e2e that `num_planes`, per-plane payloads, and block strides are populated before STORE/RETRIEVE.

The existing **high compatibility blocker also remains**: this repository still declares `LMCACHE_UPSTREAM_TAG = "v0.4.4"`, while all five newly imported 0.5 modules checked here are absent from the v0.4.4 tag. The PR description acknowledges that CI/imports fail. The core commit(s), plugin pin, and tests need to move atomically rather than merging this into the current `main` dependency set.

Static validation: the full 32-file diff passes `git diff --check`, and all added/modified Python files compile. Runtime pytest is unavailable locally because PyTorch/pytest are not installed. GitHub Code Quality currently fails on mechanical pre-commit checks; PR #298 contains the formatting follow-up but inherits these runtime blockers.

marcobarlo:
This PR now relies on https://gitcode.com/openeuler/kvcache-ops/pull/28/discuss

marcobarlo:
Results:
## Correctness and DSv4 MP vs SP bench

**Host:** 910B3-03 (8× Ascend 910B3)  
**Container:** `vllm-ascend-dsv4-lmcache` (`quay.io/ascend/vllm-ascend:v0.23.0`)  
**Date:** DSv4 MP 2026-09-14 (native packed-MLA kernel). GQA **2026-09-16**. MLA LongBench interleaved **2026-09-18** (TP=4). SP still 2026-09-11.

Dual-pass = same prompts twice (fill, then hit). A cache miss shows zero retrieves. A KV/decode bug shows retrieves plus a different answer.

| Model | Accuracy | Performance (vs SP) |
|---|---|---|
| GQA Qwen2.5-0.5B | **Pass** — long prompt 100% cache hit, needle correct both passes | n/a (tiny model) |
| GQA Qwen2.5-7B-Instruct | **Pass** — GSM8K **45/50** both, **50/50 pred match**. LongBench **38/100** both, **100/100 pred match** | No SP. GSM8K 148s→141s. LongBench 224s→221s (**1.01×**) |
| MLA DeepSeek-V2-Lite-w8a8 | **Pass (cache)** — LongBench n=100 interleaved, fill 6/100 hit 5/100. Pred match **68/100** (w8a8-Lite ramble; 256 out). No crash. | No SP. Fill 525s → hit 356s (**1.48×**). Hit **0.012 ms/tok** vs fill **0.038** |
| DSv4 Flash w8a8 | GSM8K SP **stable**; MP 49/50 vs 48/50 (**i=12** 13→12, greedy ND). LongBench MP n=16 **16/16 match**; n=100 **92/100** | Prefix_repetition **tied** (~1%). LongBench n=16 pass-2 **3.7×**; n=100 **4.4×** |

---

### Accuracy

#### GQA — Qwen2.5-0.5B-Instruct

Uniform GQA, hybrid KV **off**. Dual-pass needle (`scripts/benchmark/_mp_dual_pass.py`). TP=1 NPU 0, block 128 / chunk 256.

| Case | Prompt tokens | Store lookup req/hit | Load lookup req/hit | Needle both passes |
|---|---:|---:|---:|---|
| Short (target 64) | 106 | 0 / 0 | 0 / 0 | **correct** (below chunk) |
| Long (target 1024) | 1053 | 1024 / 0 | 1024 / **1024** | **correct** |

#### GQA — Qwen2.5-7B-Instruct

**2026-09-16.** TP=1 NPU **0**, hybrid **off**, `block_size=128`, chunk 1024. No quantization.

| Case | Fill acc | Hit acc | Pred exact match | Wall fill → hit |
|---|---:|---:|---:|---|
| GSM8K 50, prefix 8192 | 45/50 **0.90** | 45/50 **0.90** | **50/50 (1.00)** | 148s → 141s |
| LongBench n=100, mean **20 312** tok | 38/100 **0.38** | 38/100 **0.38** | **100/100 (1.00)** | 224s → 221s (**1.01×**) |

Cache-deterministic (perfect pred match). Raw: `logs/campaign/qwen7b_20260916_225711/qwen7b/`.

#### MLA — DeepSeek-V2-Lite-w8a8

**2026-09-18.** TP=4 NPUs **0–3**, hybrid **off**, `block_size=128`, chunk 1024, `QUANTIZATION=ascend`, `max_model_len=65536`, L1 20 GB. Interleaved dual-pass: for each prompt, **fill then immediately hit** (`--interleaved`) so L1 still holds that prefix. Prior batch fill-then-hit (Sep 16, TP=2) evicted earlier KV and looked like **1.02×**.

LongBench v2 n=100, prefer `any`, tokens in `[8192, 60000]`, mean **31 474** (min 10.7k, max 59.7k). `max_tokens_out=256`. No MTE, no `EngineDeadError`.

| Case | Fill acc | Hit acc | Pred exact match | Wall fill → hit |
|---|---:|---:|---:|---|
| LongBench n=100, interleaved | 6/100 **0.06** | 5/100 **0.05** | **68/100 (0.68)** | 525s → **356s** (**1.48×**) |

Task acc is the w8a8-Lite model (every request hits the 256-token cap; ~56/100 letter-extract empty). Pred flips vs fill are greedy ND on rambling decode, not a crash. Cache skip is the wall/ms-tok signal. Raw: `logs/campaign/mla_lb_interleaved_n100/`.

#### DSv4 hybrid — DeepSeek-V4-Flash-0731-w8a8

No MTP, no native prefix cache, TP=8 + EP, `block_size=32`, chunk 4096, `max_model_len=133120`, hybrid **on**. MP stores six groups (`group_tokens_per_block=[128, 4096, 32, 32, 2, 8]`).

Fixes since the Sep 11 A→D report: packed-MLA kernel (`kv_size=1`, sequential K then V); `get_tokens_per_block` always `block * compress_ratio`; host `packed_planes` is fmt 17 **only** for `element_size==1` and `hs==130` (G1 NH_CS `hs=512` uses generic 2LTD).

**GSM8K** (50 samples, temp 0, 4-shot + shared 8192-token prefix):

| Stack | Fill acc | Hit acc | Pred exact match | Date |
|---|---:|---:|---:|---|
| SP | 48/50 **0.96** | 48/50 **0.96** | **50/50** | 2026-09-11 |
| MP | 49/50 **0.98** | 48/50 **0.96** | 49/50 | **2026-09-14** |

MP mismatch: **i=12**, gold `13` (fill `#### 13`, hit `#### 12`). Greedy ND, not a systematic reload bug.

**LongBench v2** (16 short, mean 31k tokens; context-first MCQ):

| Stack | Fill acc | Hit acc | Pred exact match | Loads | Date |
|---|---:|---:|---:|---:|---|
| SP | 9/16 **0.56** | 9/16 **0.56** | 10/16 (0.625) | **16/16** | 2026-09-11 |
| MP | 10/16 **0.625** | 10/16 **0.625** | **16/16 (1.0)** | **16/16** | **2026-09-14** |

MP n=16 is **16/16 match**, including the former A→D 78k sample (`_id=6703a0ecbb02136c067cd11b`: fill A, hit A). SP still has 6/16 letter drift (acc unchanged by cancellation).

**LongBench v2 n=100** (any length, mean 54k; dual-pass only, no SP rerun):

| Stack | Fill acc | Hit acc | Pred exact match | Wall fill → hit | Date |
|---|---:|---:|---:|---|---|
| MP | 47/100 **0.47** | 46/100 **0.46** | **92/100 (0.92)** | 573s → **131s** (**4.4×**) | **2026-09-14** |

8 letter changes after reload, mixed direction, not clustered on G0 packed ids. Residual flips look like greedy first-token ND (same class as GSM8K 1/50), not the old systematic A→D.

Kernel roundtrips (`LMCache-Ascend-MP/tests/v1/test_mp_block_transfer.py`): `kg0-npu-mini`, `nh_cs-npu-1chunk`, `test_fmt17_nh_cs_hs512_does_not_use_packed_mla`, and the 78k serving retrieve all **pass**.

---

### Performance

Fair envelope (not like-for-like KV layouts). **SP benches were not rerun** on 2026-09-14.

| Item | MP | SP (`dsv4_support`) |
|---|---|---|
| Trees | `LMCache-MP` + `LMCache-Ascend-MP` | `/workspace/LMCache` @ `a7934b79` + `LMCache-Ascend` `dsv4_support` |
| Connector | `LMCacheMPConnector` | `LMCacheAscendConnector` |
| Transfer | `lmcache_driven`, L1 100 GB | in-process CPU L1 70 GB |
| Hybrid / chunk | on / 4096 | on / 4096 |
| KV geometry | six groups, `scheduler_block_size=4096` | one MLA engine, 43 layers |

#### Prefix repetition

`vllm bench serve` prefix_repetition: 32 prompts, 8 prefixes × 16384, suffix 512, out 128, conc 8. Both **32/32 success**.

| Metric | MP (Sep 14) | SP (Sep 11) | SP / MP |
|---|---:|---:|---:|
| Duration (s) | **41.84** | 41.54 | 0.99 |
| Request throughput (req/s) | **0.765** | 0.770 | 1.01 |
| Output tok/s | **97.90** | 98.62 | 1.01 |
| Total tok/s | **13024** | 13119 | 1.01 |
| Mean / median TTFT (ms) | 3984 / 3223 | **3536 / 2054** | 0.89 / 0.64 |
| Mean / median TPOT (ms) | **50.75 / 44.24** | 53.17 / 45.23 | 1.05 / 1.02 |

Throughput **tied**. Remaining gap is TTFT (prefill / retrieve / store).

#### Lognormal ISL + prefix hits (chunk-aligned)

48 prompts, min 8192, median 16384, hits = `k × 4096`. **27/48 loads**.

| Metric | MP (Sep 14) | SP (Sep 11) | SP / MP |
|---|---:|---:|---:|
| Duration (s) | **127.53** | 128.86 | 1.01 |
| Total tok/s | **9195** | 9100 | 0.99 |
| Mean / median TTFT (ms) | **4885 / 4876** | 5196 / 5008 | 1.06 / 1.03 |
| Mean / median TPOT (ms) | 91.0 / 77.7 | **87.9 / 72.9** | 0.97 / 0.94 |

#### Dual-pass wall

| Workload | Stack | Fill | Hit | Speedup |
|---|---|---:|---:|---:|
| GSM8K 50 (shared 8192 pad) | SP | 248.3s | 222.4s | — |
| GSM8K 50 | MP | **246.5s** | **246.4s** | ~1× (pad only; question tail unique) |
| LongBench n=16, mean 31k | SP | 54.3s | **18.5s** | **2.9×** |
| LongBench n=16 | MP | **53.2s** | **14.2s** | **3.7×** |
| LongBench n=100, mean 54k | MP | 573.3s | **130.9s** | **4.4×** |
| MLA V2-Lite LongBench n=100 interleaved, mean 31k | MP | 525.0s | **355.7s** | **1.48×** |
| GQA Qwen2.5-7B GSM8K 50 | MP | 148s | 141s | 1.05× |
| GQA Qwen2.5-7B LongBench n=100 | MP | 224s | 221s | 1.01× |

#### MLA V2-Lite (TP=4, interleaved)

Mean req **5.25s → 3.56s** (med **1.45×**, range 1.32–1.65). Fill `4.06s + 0.038 ms/tok`; hit `3.19s + 0.012 ms/tok`. Decode still dominates: **256 completion tokens** every request (~3.2s fixed). Longer prefixes skip more:

| Prompt | n | Mean tok | Speedup |
|---|---:|---:|---:|
| &lt;20k | 28 | 16 846 | **1.40×** |
| 20–40k | 43 | 29 019 | **1.45×** |
| &gt;40k | 29 | 49 239 | **1.57×** |

DSv4 LongBench is the workload where LMCache cuts most of the wall (1-token letter). Prefix-padded GSM8K / GQA 7B stay ~1× because the unique tail (or decode) is most of the compute.

#### Load wall (why MP TTFT still lags)

SP ranks run in parallel. MP is one server: 8 ranks serialize. Request load wall ≈ sum of 8 rank ops.

| Workload | SP request wall | MP request wall | Per-rank (SP / MP) |
|---|---|---|---|
| prefix_repetition | **9.3 ms** (parallel) | **~82 ms** (serial) | 9.3 / **10.3** ms |
| lognormal v2 | **8.1 ms** | **~98 ms** | 7.7 / **12.2** ms |
| GSM8K | **6.2 ms** | **~55 ms** | 5.8 / **6.9** ms |
| LongBench v2 hit | **16 ms** | **~96 ms** | 16.0 / **12.0** ms |

Per-rank kernel is now close to (or faster than) SP. The remaining TTFT gap is **rank serialization** on the retrieve path (~9× on GSM8K 8192). Fast MP stores are ~10 ms (copy-dominated); long SP `cost` lines include `wait_for_forward`.

---

### Reproduce

```bash
# MLA LongBench interleaved fill→hit (NPUs 0-3 TP=4 :8019)
python3 scripts/benchmark/run_longbench_v2_dual.py \
  --model mla --tokenizer /workspace/models/DeepSeek-V2-Lite-w8a8/tokenizer.json \
  --chat-url http://127.0.0.1:8019/v1/chat/completions \
  --limit 100 --min-tokens 8192 --max-tokens 60000 --prefer-length any \
  --max-tokens-out 256 --interleaved \
  --out-dir logs/campaign/mla_lb_interleaved_n100

# GQA 0.5B needle
docker exec -u root -e MODEL_PATH=/workspace/models/Qwen2.5-0.5B-Instruct \
  -e MODEL_NAME=qwen05b vllm-ascend-dsv4-lmcache \
  bash /workspace/dsv4-serving/scripts/serving/single_node_mp_small.sh launch
docker exec -u root vllm-ascend-dsv4-lmcache python3 \
  /workspace/dsv4-serving/scripts/benchmark/_mp_dual_pass.py \
  --model qwen05b --tokenizer /workspace/models/Qwen2.5-0.5B-Instruct/tokenizer.json \
  --chunk 256 --short 64 --long 1024

# GQA 7B GSM8K+LongBench
./scripts/benchmark/run_simple_quality_campaign.sh

# DSv4 MP bench
./scripts/run_mp_mode.sh launch
LABEL=mp PORT=8019 ./scripts/benchmark/run_bench_dsv4_compare.sh

# GSM8K fill then hit
python3 scripts/benchmark/run_gsm8k_dual.py \
  --model dsv4 --tokenizer /workspace/models/DeepSeek-V4-Flash-0731-w8a8/tokenizer.json \
  --chat-url http://127.0.0.1:8019/v1/chat/completions \
  --limit 50 --prefix-tokens 8192 --out-dir logs/bench/gsm8k_mp

# LongBench v2 n=16 / n=100
python3 scripts/benchmark/run_longbench_v2_dual.py \
  --model dsv4 --tokenizer /workspace/models/DeepSeek-V4-Flash-0731-w8a8/tokenizer.json \
  --chat-url http://127.0.0.1:8019/v1/chat/completions \
  --limit 16 --min-tokens 8192 --max-tokens 120000 --prefer-length short \
  --out-dir logs/bench/longbench_mp
```

Do **not** export `PYTHONPATH=.../logs/overlay-vllm-ascend`. Logical block size for DSv4 MP comes from the LMCache-Ascend-MP plugin patch, not a vLLM-Ascend source edit.

matthewygf:
Rechecked the three new commits at `a53b1e6`. The production plane-geometry gap from my previous comment is now addressed: `install_plane_geometry_fill()` wraps the upstream descriptor factory for formats 16/17, fills payload/stride bytes from the real tensors, and the new mixed-plane kernel cases exercise the resulting geometries.

Two **high-severity integration blockers remain**, and one is newly introduced by the latest cleanup:

1. **The committed tree no longer contains the block kernel it includes.** `csrc/mp_mem_kernels.cpp` now includes `kernels/multi_layer/multi_layer_block_mem_kernels.h`, while commits `70c7f04`/`a53b1e6` delete the vendored `new_kernels/` implementation and its CMake copy. However, the `third_party/kvcache-ops` gitlink did **not** advance: both base `e05a757` and current head point to `de43ed5d5fa4f25e87c7a86272b4759dd409af03`. That pinned tree contains the existing `multi_layer_mem_kernels*` files but not `multi_layer_block_mem_kernels.h`, `block_transfer_layout.h`, or `multi_layer_block_transfer_kernel`. A clean recursive checkout therefore fails at compilation; the contributor's local checkout of kvcache-ops PR #28 is not represented in this PR. Please land/pin the exact kvcache-ops commit first (or retain the vendored implementation) and validate from a fresh recursive clone.

2. **The exact referenced LMCache dependency still cannot construct the NPU cache context.** LMCache#5138 has moved to `218cd67`, but `NpuDeviceSpec` still has no `create_cache_context()` override; the inherited method raises `NotImplementedError`. This branch's e2e still calls `NpuDeviceSpec().create_cache_context(...)` directly, so it cannot reach the descriptor wrapper or kernel against the named official dependency graph. The reported hardware runs use the separate `MP_dsv4_3_split` forks. Please reference/land the missing core NPU context implementation and rerun the e2e against the exact committed LMCache, plugin, and kvcache-ops revisions.

The acknowledged **v0.4.4 compatibility blocker** also remains: `LMCACHE_UPSTREAM_TAG` is unchanged while this patch imports 0.5-only modules. These dependency changes need to land atomically.

Static validation: the complete 30-file diff passes `git diff --check`; all 17 added/modified Python files compile. Current Code Quality fails on ten `ruff` E501 violations. Runtime tests and the NPU build are unavailable locally, but the missing header is established directly from the committed gitlink tree.

lmcache_ascend/__init__.py
```diff
@@ -25,6 +29,198 @@ def _is_vllm_runtime():
     return "vllm" in sys.modules or any("vllm" in arg for arg in sys.argv)
 
 
+def _seed_partial_lmcache_device_ops() -> None:
```
marcobarlo, line 32: @chloroethylene  Can we get rid of this?

lmcache_ascend/__init__.py
```diff
@@ -526,6 +781,14 @@ def handle_preemptions(self, preempted_req_ids):
     vllm_lmcache_connector.LMCacheConnectorV1.handle_preemptions = handle_preemptions
 
 
+def _patch_make_page_buffer_shape_desc():
```
marcobarlo: we have to get rid of this.

matthewygf:
Rechecked the three new commits through `20b0f51`.

The new active-work launch sizing, skip-all early return, runtime UB query, and consolidated event-recorder tests do not resolve the existing integration blockers:

1. **High — a clean recursive checkout still cannot build the new kernel.** [`mp_mem_kernels.cpp` still includes `multi_layer_block_mem_kernels.h`](https://github.com/LMCache/LMCache-Ascend/blob/20b0f51caf8781dbfa808f2b281c2d6ba2505840/csrc/mp_mem_kernels.cpp#L19-L22), while the committed [`third_party/kvcache-ops` gitlink](https://github.com/LMCache/LMCache-Ascend/tree/20b0f51caf8781dbfa808f2b281c2d6ba2505840/third_party/kvcache-ops) remains `de43ed5d5fa4f25e87c7a86272b4759dd409af03`. That revision has neither this header nor the block-transfer kernel symbol. The external kvcache-ops PR #28 must be represented by an updated gitlink (or the implementation restored) before this branch is self-contained.

2. **High — the exact referenced LMCache path still stops before transfer.** LMCache PR #5138 now points to `218cd67`, but `NpuDeviceSpec` still has no `create_cache_context()` override; the inherited factory raises `NotImplementedError`. This PR's E2E continues to call that missing override at [`test_npu_lmcache_driven_e2e.py`](https://github.com/LMCache/LMCache-Ascend/blob/20b0f51caf8781dbfa808f2b281c2d6ba2505840/tests/v1/multiprocess/test_npu_lmcache_driven_e2e.py#L216-L225).

3. **High compatibility blocker unchanged:** [`LMCACHE_UPSTREAM_TAG` is still `v0.4.4`](https://github.com/LMCache/LMCache-Ascend/blob/20b0f51caf8781dbfa808f2b281c2d6ba2505840/lmcache_ascend/__init__.py#L15-L18), while this implementation requires the 0.5 MP modules. The plugin/core pin and exact-dependency tests need to move together.

Static validation: all 17 added/modified Python files compile and `git diff --check` passes. [Code Quality fails](https://github.com/LMCache/LMCache-Ascend/actions/runs/35623684339) because isort, Ruff/format, codespell, and clang-format rewrite or reject the current tree. Runtime NPU tests and a clean CANN build are unavailable locally. Recommendation remains **do not merge** until the dependency revisions are committed and the E2E is run against those exact revisions.

marcobarlo:
Setting as draft until we remove the monkey patching from make_page_buffer_shape_desc() and find a clean design for this

